PR #309, opened 2014-07-20
reload modules with F6 - fix #285

### LINKED ISSUE #285: Reevaluate key
There should be some way to re-run an entire bpython session, for the use case of editing a module and playing with it interactively in bpython at the same time. I'm not sure how to implement this, but the cleanest way would be restarting bpython and running it again with --type tempfile-with-text-from-prev-session-in-it.py

An alternative would be clearing out sys.modules to make sure things get reloaded, but I think that has edge cases.

---
- Bitbucket: https://bitbucket.org/bobf/bpython/issue/285
- Originally Reported By: [ ](http://bitbucket.org/thomasballinger)
- Originally Created At: 2014-05-15T19:39:19.670

thomasballinger:
Key that clears the `sys.modules` cache of modules and reevaluates session.

### DIFF
--- a/bpython/config.py
+++ b/bpython/config.py
@@ -80,6 +80,7 @@ def loadini(struct, configfile):
             'show_source': 'F2',
             'suspend': 'C-z',
             'undo': 'C-r',
+            'reimport': 'F6',
             'search': 'C-o',
             'up_one_line': 'C-p',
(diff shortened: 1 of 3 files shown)


kevinisaac:
Good feature! +1

PR #365, opened 2014-08-28
Reverse incremental search

thomasballinger:
ctrl-r and ctrl-s working approximately like the do in bash - reverse incremental search and incremental search.

This feels alright to me, but I don't use ctrl-r much normally, so would appreciate someone that cares more about these shortcuts trying them out.

As of right now they're hardcoded to ctrl-r and ctrl-s, but I'll make them configs soon.


karolyi:
So, how can I enable this on OSX? `Esc-r` doesn't seem to work in Yosemite terminal.

thomasballinger:
There's an option in the the OSX terminal to send "Esc+" for option, or if there isn't there's one in iterm.

PR #388, opened 2014-09-12
fix control-c? (Seems to work, could use review)

thomasballinger:
I'm not confident about this fix - it looks like maybe this functionality used to work, but now the KeyboardInterrupt isn't raised anymore for some reason?

### DIFF
--- a/bpython/cli.py
+++ b/bpython/cli.py
@@ -999,6 +999,9 @@ class CLIRepl(repl.Repl):
         elif key == '\x18':
             return self.send_current_line_to_editor()
 
+        elif key == '\x03':
+            raise KeyboardInterrupt()
+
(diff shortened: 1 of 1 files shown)


snakecharmer1024:
control-c doesn't work for me on OSX

thomasballinger:
What terminal are you using, in what situation does ctrl-c not work?

snakecharmer1024:
I have some variable, say v and I type into OSX Terminal (comes with OSX) 

v

Then if it ends up being a gigantic datapoint, with maybe 10M elements, I try to stop it from printing out everything with ctrl-c, but pressing repeatedly or holding it does not stop the torrent of data that can go on for a long time. I usually end up exiting the terminal when it happens.

Issue #432, opened 2015-01-07
bpython-curtsies crashes if no readline history on enter


thomasballinger:
I think this is if you don't have a ~/.pythonhist file.

thomasballinger:
~~~.py
Traceback (most recent call last):
  File "/usr/local/bin/bpython-curtsies", line 9, in <module>
    load_entry_point('bpython==0.13.1', 'console_scripts', 'bpython-curtsies')()
  File "/usr/local/lib/python2.7/site-packages/bpython/curtsies.py", line 60, in main
    mainloop(config, locals_, banner, interp, paste)
  File "/usr/local/lib/python2.7/site-packages/bpython/curtsies.py", line 92, in mainloop
    process_event(term.get_annotated_event(idle=find_iterator))
  File "/usr/local/lib/python2.7/site-packages/bpython/curtsies.py", line 76, in process_event
    repl.process_event(e)
  File "/usr/local/lib/python2.7/site-packages/bpython/curtsiesfrontend/repl.py", line 341, in process_event
    self.on_enter()
  File "/usr/local/lib/python2.7/site-packages/bpython/curtsiesfrontend/repl.py", line 374, in on_enter
    self.rl_history.last()
  File "/usr/local/lib/python2.7/site-packages/bpython/repl.py", line 241, in last
    return self.entries[0]
IndexError: list index out of range
~~~

thomasballinger:
would also be good to check if this happens with bpython as well

sebastinas:
It does not happen with bpython.cli.

sebastinas:
Now fixed.

PR #433, opened 2015-01-08
fix bpython-curtsies crashes if no readline history on enter #432

lvgelder:
It actually works correctly if there is no existing history the first time you press enter. But then insert_into_history sets entries to [] and tries to read from the history file, which will do nothing if the history file is empty.

This is fixed properly on master, for now just change last() to return an empty string if history entries are empty.

### DIFF
--- a/bpython/repl.py
+++ b/bpython/repl.py
@@ -238,7 +238,7 @@ class History(object):
         """Move forward to the end of the history."""
         if not self.is_at_start:
             self.index = 0
-        return self.entries[0]
+        return self.entries[0] if self.entries else ''
 
     @property
     def is_at_end(self):
--- a/bpython/test/test_repl.py
+++ b/bpython/test/test_repl.py
@@ -75,6 +75,13 @@ class TestHistory(unittest.TestCase):
         self.assertTrue(self.history.is_at_start)
         self.assertFalse(self.history.is_at_end)
 
+    def test_last_with_no_history_doesnt_throw_error(self):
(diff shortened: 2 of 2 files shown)


sebastinas:
I don't think this is the correct approach to fix that issue. If the history is empty, `last` should throw because there is no last element. The caller of this function needs to handle this.

As far as I can tell, `insert_into_history` will always keep a history with atleast one entry. `rl_history` starts with one entry (`''`). In `insert_into_history`, `entries` is set to `[]`, then entries from the file are loaded, and after that, a new entry is appended. Again at least one entry. I think we actually need to find the place where `entries` is reduced to an empty list.

thomasballinger:
@lvgelder found that entries is set to [] when an empty file is loaded:
~~~.py
$ bpython-curtsies
>>> from bpython.repl import History
>>> h = History()
>>> h.entries
['']
>>> h.save('tmphist', 'utf8')
>>> h.load('tmphist', 'utf8')
>>> h.entries
[]
~~~
because the `History.append` (called by `History.load`) doesn't add empty lines. We could add a special case to `load` for this if always having entries is an intended invariant, for the record @lvgelder had a more comprehensive fix but I suggested making as few changes as possible for the bugfix, so my fault :)

thomasballinger:
@sebastinas and I looked into this and the reason is bpython-curtsies does a `rl_history.last()` where it should do a `rl_history.reset()`

sebastinas:
Thanks, merged.

PR #446, opened 2015-01-17
Add deactivate message when turning off auto-reload

### LINKED ISSUE #442: No message for turning auto-reloading off
Maybe it's not getting turned off, or maybe there's just no notification?

The text also seems not to expire by itself, requires hitting a key for it to disappear.

keyanP:
Fixes #442


thomasballinger:
Thanks, this is great!

PR #451, opened 2015-01-21
Refactor and fix keybinding issue (fixes #447)

### LINKED ISSUE #447: problem with keybindings
With
```
[keyboard]
help = F2
show_source = F9
```
F2 is stilled mapped to `show_source`.

keyanp:
This changes the logic for the function `get_key_no_doublebind()` and removes the mutable list which was previously used to keep track of keybindings which were already assigned. Although this fix passes the test cases and the specific case mentioned in #447, the logic is such that specifying the same custom keybinding for two commands will result in only one command being bound. Also I am getting 12 skipped tests when running nosetests, any ideas why this is happening?

With:
`help = F9`
`show_source = F9`

F9  is mapped to `show_source` and `help` is not mapped to any key. I think this is because the keybindings are assigned sequentially so because `show_source` is assigned first (line 182) vs `help` (line 210).

I think this is still a substantial improvement over the old code because it is easier to understand and has more functionality than before. Although the problem of correctly handling custom keybinding is not completely resolved. Knowing the exact desired functionality would be helpful though.

### DIFF
--- a/bpython/config.py
+++ b/bpython/config.py
@@ -74,7 +74,6 @@ def loadini(struct, configfile):
             'transpose_chars': 'C-t',
             'clear_line': 'C-u',
             'clear_screen': 'C-l',
-            'kill_line': 'C-k',
             'clear_word': 'C-w',
             'cut_to_buffer': 'C-k',
             'delete': 'C-d',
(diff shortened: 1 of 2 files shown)


keyanp:
Thanks Sebastian, so I'm having an issue passing the `test_keybindings_default_contains_no_duplicates()` test for some reason. But I'm working on it. Please ignore the messy branch management, I'm still getting the hang of that.

keyanp:
Fixed the merge conflicts and also passes the new tests. The issue was that `kill_line` and `cut_to_buffer` had the same default key assigned, their functionality was also the same. I removed `kill_line` and changed manual_readline.py to reflect this.

Issue #455, opened 2015-01-22
Crash on dictionary completion of numpy array


thomasballinger:
~~~.py
(bpython)tom-mba:bpython tomb$ bpython
bpython version 0.13-416 on top of Python 2.7.5 /Users/tomb/.virtualenvs/bpython/bin/python2.7
>>> import numpy
>>> a = numpy.zeros((2, 3))
>>> a
array([[ 0.,  0.,  0.],
       [ 0.,  0.,  0.]])
>>> a
Traceback (most recent call last):
  File "/Users/tomb/.virtualenvs/bpython/bin/bpython", line 9, in <module>
    load_entry_point('bpython==0.13-416', 'console_scripts', 'bpython')()
  File "/Users/tomb/Dropbox/code/bpython/bpython/curtsies.py", line 69, in main
    mainloop(config, locals_, banner, interp, paste, interactive=(not exec_args))
  File "/Users/tomb/Dropbox/code/bpython/bpython/curtsies.py", line 151, in mainloop
    process_event(e)
  File "/Users/tomb/Dropbox/code/bpython/bpython/curtsies.py", line 114, in process_event
    repl.process_event(e)
  File "/Users/tomb/Dropbox/code/bpython/bpython/curtsiesfrontend/repl.py", line 433, in process_event
    return self.process_key_event(e)
  File "/Users/tomb/Dropbox/code/bpython/bpython/curtsiesfrontend/repl.py", line 562, in process_key_event
    self.add_normal_character(e)
  File "/Users/tomb/Dropbox/code/bpython/bpython/curtsiesfrontend/repl.py", line 767, in add_normal_character
    self.cursor_offset += 1
  File "/Users/tomb/Dropbox/code/bpython/bpython/curtsiesfrontend/repl.py", line 1253, in _set_cursor_offset
    self.update_completion()
  File "/Users/tomb/Dropbox/code/bpython/bpython/curtsiesfrontend/repl.py", line 796, in update_completion
    self.list_win_visible = BpythonRepl.complete(self, tab)
  File "/Users/tomb/Dropbox/code/bpython/bpython/repl.py", line 537, in complete
    complete_magic_methods=self.config.complete_magic_methods)
  File "/Users/tomb/Dropbox/code/bpython/bpython/autocomplete.py", line 348, in get_completer_bpython
    **kwargs)
  File "/Users/tomb/Dropbox/code/bpython/bpython/autocomplete.py", line 330, in get_completer
[...]

PR #457, opened 2015-01-22
Add tests for dictionary key completion

maryakins:
One of the test is failing due to #455

### DIFF
--- a/bpython/test/test_autocomplete.py
+++ b/bpython/test/test_autocomplete.py
@@ -145,3 +145,38 @@ class TestFilenameCompletion(unittest.TestCase):
         self.assertEqual(self.completer.format('/hello/there/'), 'there/')
         self.assertEqual(self.completer.format('/hello/there'), 'there')
 
+class MockNumPy(object):
+    """
+        This is a mock numpy object that raises an error when it is trying to be converted to a boolean.
+    """
+    def __nonzero__(self):
+        raise ValueError("The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()")
+
+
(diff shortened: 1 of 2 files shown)


bpython/test/test_autocomplete.py
```diff
@@ -145,3 +145,38 @@ def test_formatting_takes_just_last_part(self):
...
+        result=com.matches(2,"l[" , local)
+        self.assertSetEqual(result,set())
+
+    def test_obj_that_does_not_allow_bool_tests(self):
+        com = autocomplete.DictKeyCompletion()
+        local={'mNumPy':MockNumPy(),}
+        try:
+            com.matches(7,"mNumPy[" , local)
+        except ValueError:
+            raise AssertionError("Dict key completion raised value error.")
```
sebastinas: Just let the `ValueError` pass. If a test throws anything, it is considered as failure. I'd prefer a `self.asserSetEqual(...)` as above.

maryakins:
Also another test revealed a bug in 2.6

thomasballinger:
Nosetests reports an error for me instead of a failure if I let the ValueError pass - do you mean we ought to consider errors failures and not worry about catching these?

sebastinas:
Ah yes, it will be reported as error instead of failure. But as said on IRC, the expected output of that `matches` call is probably an empty set and that is what we should test for.

thomasballinger:
Ah yeah testing for the expected behavior vs a bug definitely makes sense

sebastinas:
Thank you, @maryakins.

PR #467, opened 2015-01-28
French translation

tarekziade:


tarekziade:
One thing I am curious about is non-ascii characters in terminals. A lot of french people are using plain non-utf8 terminals and non-ascii characters look weird. A good strategy could be to drop them, it's not an issue for understanding the text in French. 

Let me know.

sebastinas:
What characters do you mean? The border of the completion box? One can disable them in the config.

Issue #34, opened 2014-07-04
bpython could pprint by default


ikanobori:
It would be nice if bpython could pprint the output by default. In an interactive interpreter, pretty printing things is much better, than the default printing.

Of course, this could be configurable for people that don't like pretty things. :)

---
- Bitbucket: https://bitbucket.org/bobf/bpython/issue/34
- Originally Reported By: [ ](http://bitbucket.org/nilton)
- Originally Created At: 2009-06-15T03:25:32.728

vdboor:
Sadly I couldn't easily get this to work. I don't see why BPython can't offer a convienience setting for this, instead of having to work through Python Internals. IPython does offer it:

```
$ ipython help notebook | grep pprint -A1
--pprint
    Enable auto pretty printing of results.
--
--no-pprint
    Disable auto auto pretty printing of results.
```

PR #485, opened 2015-02-18
Add feature to enable debugging on uncaught exceptions.

kdart:
The debugger is selected by an environment variable, PYTHON_DEBUGGER. This
should be set to a module path. If it's not set it will use the standard pdb.

Press F12 to toggle the feature off and on. When off the original behaviour is
used.

Also add pretty-printing option. Currently this is only selected when invoked
from the CLI.

### DIFF
--- a/bpython/cli.py
+++ b/bpython/cli.py
@@ -49,6 +49,7 @@ import math
 import re
 import time
 import functools
+from pprint import pprint
 
 import struct
 if platform.system() != 'Windows':
@@ -160,6 +161,7 @@ class FakeStdin(object):
         self.encoding = getpreferredencoding()
         self.interface = interface
         self.buffer = list()
+        self.errors = "strict"
 
     def __iter__(self):
         return iter(self.readlines())
@@ -957,6 +959,13 @@ class CLIRepl(repl.Repl):
             self.write2file()
             return ''
 
+        elif key in key_dispatch[config.debug_key]:
+            if sys.excepthook is not debugger_hook:
+                sys.excepthook = debugger_hook
+            else:
+                sys.excepthook = sys.__excepthook__
+            return ''
+
         elif key in key_dispatch[config.pastebin_key]:
             self.pastebin()
             return ''
@@ -1696,7 +1705,8 @@ def init_wins(scr, config):
         (_('Save'), config.save_key),
         (_('Pastebin'), config.pastebin_key),
         (_('Pager'), config.last_output_key),
-        (_('Show Source'), config.show_source_key)
+        (_('Show Source'), config.show_source_key),
+        (_('Auto Debug'), config.debug_key),
     )
 
     message = '  '.join('<%s> %s' % (key, command) for command, key in commands
@@ -1944,9 +1954,46 @@ def main_curses(scr, args, config, interactive=True, locals_=None,
     return (exit_value, clirepl.getstdout())
 
 
+def prettydisplayhook(obj):
+    pprint(obj)
+
+
+def debugger_hook(exc, value, tb):
+
+    if exc in (SyntaxError, IndentationError, KeyboardInterrupt):
+        sys.__excepthook__(exc, value, tb)
+        return
+
+    global stdscr
+    from bpython import debugger
+
+    orig_stdin = sys.stdin
+    orig_stdout = sys.stdout
+    orig_stderr = sys.stderr
+
+    sys.stdin = sys.__stdin__
+    sys.stderr = sys.__stderr__
+    sys.stdout = sys.__stdout__
+
+    stdscr.keypad(0)
+    curses.echo()
+    curses.nocbreak()
+    curses.endwin()
+    try:
+        debugger.post_mortem(tb, exc, value)
+    finally:
+        sys.stdin = orig_stdin
+        sys.stdout = orig_stdout
+        sys.stderr = orig_stderr
+        curses.cbreak()
+        curses.noecho()
+        stdscr.keypad(1)
+        if stdscr is None:
+            stdscr = curses.initscr()
+
+
 def main(args=None, locals_=None, banner=None):
     translations.init()
-
 
     config, options, exec_args = bpython.args.parse(args)
 
@@ -1954,6 +2001,11 @@ def main(args=None, locals_=None, banner=None):
     orig_stdin = sys.stdin
     orig_stdout = sys.stdout
     orig_stderr = sys.stderr
+
+    if options.debugger:
+        sys.excepthook = debugger_hook
+    if options.pretty:
+        sys.displayhook = prettydisplayhook
 
     try:
         (exit_value, output) = curses_wrapper(
--- a/bpython/cli.py
+++ b/bpython/cli.py
@@ -81,6 +81,12 @@ from bpython.keys import cli_key_dispatch as key_dispatch
 from bpython import translations
 from bpython.translations import _
 
+# Debugging support
+try:
+    from bpython import debugger
+except ImportError as err:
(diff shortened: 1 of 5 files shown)


sebastinas:
Thank you for implementing this feature. I'll add some comments to the code once I've had the time to play with it a bit.

The test suite now fails. This needs to be fixed.

I'm afraid pretty printing won't make it into bpython. See #34 for details. But I'm happy to merge bits for the documentation explaining the use of PYTHONSTARTUP.

bpython/config.py
```diff
@@ -119,7 +119,8 @@ def loadini(struct, configfile):
             'transpose_chars': 'C-t',
             'undo': 'C-r',
             'up_one_line': 'C-p',
-            'yank_from_buffer': 'C-y'
+            'yank_from_buffer': 'C-y',
+            'debug': 'F12',
```
sebastinas: Please keep the entries sorted by key.

bpython/cli.py
```diff
@@ -160,6 +161,7 @@ def __init__(self, interface):
         self.encoding = getpreferredencoding()
         self.interface = interface
         self.buffer = list()
+        self.errors = "strict"
```
sebastinas: Why? Does this fix a bug? If so, this should be a separate commit in a separate pull request.

bpython/debugger.py
```diff
@@ -0,0 +1,36 @@
...
+
+import os
+import sys
+
+
+post_mortem = None
+
+def _get_debugger():
+    global post_mortem
+    modname = os.environ.get("PYTHON_DEBUGGER", "pdb")
```
sebastinas: This should default to bpdb.

bpython/debugger.py
```diff
@@ -0,0 +1,36 @@
...
+from __future__ import division
+
+
+import os
+import sys
+
+
+post_mortem = None
+
+def _get_debugger():
```
sebastinas: The whole function lacks proper error handling. What if a module named `modname` does not exist? What if `mod.pm` does not exist?

bpython/debugger.py
```diff
@@ -0,0 +1,36 @@
...
+
+"""
+Debugger factory. Set PYTHON_DEBUGGER to a module path that has a post_mortem
+function in it. Defaults to pdb. This allows alternate debuggers to be used,
+such as pycopia.debugger. :)
+"""
+from __future__ import absolute_import
+from __future__ import print_function
+from __future__ import unicode_literals
+from __future__ import division
```
sebastinas: It looks like they are not need here.

bpython/cli.py
kdart: I forget now... I actually made these original changes a long time ago. I think was related to getting it to work with Python 3. It could probably be removed, but I kept it just in case there was still a reason for it.

bpython/debugger.py
kdart: It will raise an ImportError so that the user (who presumably knows what he/she is doing if they are going to set PYTHON_DEBUGGER) can fix it. 

Only "post_mortem" is looked for.  "pm" is a local variable. 

Sometimes it's best to just let errors pass up, so the user can get a full stack trace and know where to fix it.

bpython/debugger.py
kdart: Standard header for my code, for better Python 3 compatibility.

kdart:
Well, it worked for me. :)  What is the error?

bpython/debugger.py
kdart: Ok. But that one doesn't always get back to the original bpython, and can also fail if you don't have all the dependencies (which I don't have because the curses variant doesn't need them).

bpython/debugger.py
sebastinas: No, crashing because the user typoed PYTHON_DEBUGGER is not an option. The whole session would be lost.

kdart: You mean if the typoed the value of PYTHON_DEBUGGER.  They will lose their session once and then fix the problem.  However that can be easily fixed I guess and write out an error message.

kdart: Fixed now. The echo method is used to print a warning if the debugger module couldn't be imported earlier.

kdart:
Any other issues with this request?

sebastinas:
Support in the curtsies frontend is missing.

bpython/cli.py
sebastinas: It should still be a separate commit. Please remove it. It has nothing to do with the new feature.

thomasballinger:
This is a cool feature! I'm trying to think about how it should work with the curtsies frontend.

thomasballinger:
UI-wise, it'd be nice to have messages similar to when auto-reloading is turned on.

There are some unicode issues atm, but fixing these seems doable.

I wonder if the debugging should happen in an alternate screen (so history of it is erased) or inline with the rest of the session?

thomasballinger:
I've got this nominally working in the Curtsies frontend, but going back into bpython doesn't work - the Curtsies frontent can't be launched from within itself, it doesn't like the fake stdin/out/err. A workaround would be to embed with the curses frontend, but that doesn't seem great for going forward. This is a problem with bpdb and Curses though, not this PR.

See what you think of https://github.com/bpython/bpython/tree/debugging-uncaught-exceptions, it's not probably ready yet but I'd like to get it up to snuff and merge without requiring getting launching bpython from bpdb working, that should be another bug.

kdart:
Sorry for the delay. I haven't used Curtsies before so I needed some time to look into it.  Frankly, still not sure I like it. The idea seems OK, though. But is it really needed? The curses UI is still faster, and it works.

sebastinas:
The curses UI will be removed sooner or later. It is no longer maintained.

thomasballinger:
Blocked on https://github.com/bpython/bpython/issues/520, once in bpdb bpython-curtsies can't be run yet. kdart make with work for bpython-curses with https://github.com/bpython/bpython/pull/485/files#diff-15c8a80fae5902b7d458888b6149c023R1967, I think I can figure something similar out.

thomasballinger:
I've tried to allow multiple bpython sessions by making the terminal changing objects global, see https://github.com/bpython/bpython/compare/debugging-uncaught-exceptions

I'm not sure how I feel about this. It's a hack, and history in the new bpython session isn't preserved. I'd also like to have the debugger turned off once this new bpython session starts - I think it gets confusing when the second (debugging) bpython session has an error and a second pdb is opened up.

This may be too complicated to be worth it - better might be to disallow bpython from being called, and if a session is already going on making the B command in bpdb a nop.

I'd also prefer a solution that would allow [pudb](https://pypi.python.org/pypi/pudb) to be used as a debugger, but I think that would require using a ptty or something as drastic.

thomasballinger:
@kdart Any advice on getting pycopia working? I'm trying to run it locally to test that the debugger works but having trouble with pyrex versions (I think).

thomasballinger:
I'd be in favor of taking this PR and making pdb the default debugger for now, and pushing off getting bpython working in bpython. This would mean making the Curtsies objects global, just doing something like https://github.com/bpython/bpython/compare/simple-auto-debugging

kdart:
@thomasballinger Hello. Thanks for checking it out. The setuptools package used to automate the building of pyx files using pyrex or cython. But now it doesn't seem to do that. You can do it manually first, for now.

	$ cd src/pycopia/utils
	$ cython pycopia.itimer.pyx
	$ python2.7 setup.py build

The build and install should work fine once that is done.

richin13:
Is this ever going to make it? :cry:

PR #508, opened 2015-03-13
Add Ctrl-e to complete current suggestion.

amjith:
This PR adds Ctrl-e to select the current suggestion in the curtsies frontend.

I love how the curtsies front-end shows completions from the history as I type. In order to complete that current suggestion I have to press 'Right-Arrow' or 'Ctrl-f'. I presume this is inspired by fish shell. Fish shell also allows Ctrl-e to select the current suggestion and this PR adds that behavior to bpython.

### DIFF
--- a/bpython/curtsiesfrontend/repl.py
+++ b/bpython/curtsiesfrontend/repl.py
@@ -594,12 +594,11 @@ class Repl(BpythonRepl):
         if self.stdin.has_focus:
             return self.stdin.process_event(e)
 
-        if (e in ("<RIGHT>", '<Ctrl-f>') and
+        if (e in ("<RIGHT>", '<Ctrl-f>', '<Ctrl-e>') and
                 self.config.curtsies_right_arrow_completion and
                 self.cursor_offset == len(self.current_line)):
             self.current_line += self.current_suggestion
             self.cursor_offset = len(self.current_line)
-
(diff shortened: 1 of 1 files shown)


sebastinas:
This should move to the config (ideally supporting multiple key bindings for a function). Hardcoding this in the curtsies frontend is not nice.

thomasballinger:
Some thing like
~~~
diff --git a/bpython/curtsiesfrontend/repl.py b/bpython/curtsiesfrontend/repl.py
index cb913b7..632fe72 100644
--- a/bpython/curtsiesfrontend/repl.py
+++ b/bpython/curtsiesfrontend/repl.py
@@ -594,7 +594,9 @@ class Repl(BpythonRepl):
         if self.stdin.has_focus:
             return self.stdin.process_event(e)

-        if (e in ("<RIGHT>", '<Ctrl-f>') and
+        if (e in (key_dispatch[self.config.right_key] +
+                  key_dispatch[self.config.end_of_line_key] +
+                  ("<RIGHT>",)) and
                 self.config.curtsies_right_arrow_completion and
                 self.cursor_offset == len(self.current_line)):
             self.current_line += self.current_suggestion
~~~

thomasballinger:
@sebastinas The current setting in the config for this is "right_arrow_completion" which currently enables this behavior for right arrow and ctrl-f.

Issue #336, opened 2014-08-11
Other completion modes


thomasballinger:
@jasonLaster wrote code for fuzzy matching and substring matching. It'd be cool to use this somehow. He implemented it with a config option where you chose which you wanted - I wonder if there's something even better. If not, it would be nice to add these back. (they were removed when I "refactored" completion).

thomasballinger:
Completion modes are back, see #510

jasonLaster:
This is a good day for poor typers and short term memory loss! 

Seriously, great to see productivity wins.

PR #510, opened 2015-03-15
completion matches in the middle of words

Leryan:
May be related to issue #336.

It gives matches without the need to start with typed chars.

For exemple:

```
a = dict()
a.op   # will match "copy" and "pop"
```

I find strange that autocomplete_mode just doesn't work. Can someone explain me what is the purpose of this option now?

### DIFF
--- a/bpython/autocomplete.py
+++ b/bpython/autocomplete.py
@@ -557,4 +557,4 @@ def _callable_postfix(value, word):
 
 
 def method_match(word, size, text):
-    return word[:size] == text
+    return text in word


sebastinas:
The other completion modes got removed while rewriting the auto completion. We should add them back.

Once the code to have different completion modes is back, we could talk add one that is stricter, but at this point the default completion mode is not going to change.

sebastinas:
Completion modes are back. Does this solve your problem?

Issue #229, opened 2014-07-04
patch inspect.getsource so that I can get the source of interactively defined functions


ikanobori:
It seems like bpython is already saving a history of everything I type (so that it can be printed on exit); it would be really nifty if 'inspect.getsource' could show me the source of a function so that I could save things out individually.

---------------------------------------
- Bitbucket: https://bitbucket.org/bobf/bpython/issue/229
- Originally Reported By: [ ](http://bitbucket.org/glyph)
- Originally Created At: 2012-10-04T02:18:02.077

thomasballinger:
Looks like ipython does this by compiling functions with unique filenames for each prompt.

PR #515, opened 2015-03-25
Remember the source of interactively defined functions

rhymeswithcycle:
- Patches the linecache module to achieve this
- inspect.getsource() works for functions defined on the console
- tracebacks show code for frames inside interactively defined functions

fixes bpython/bpython#229

### DIFF
--- a/bpython/history.py
+++ b/bpython/history.py
@@ -25,7 +25,6 @@
 
 from __future__ import unicode_literals
 import io
-import linecache
 import os
 import stat
 from itertools import islice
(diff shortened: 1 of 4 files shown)


bpython/history.py
```diff
@@ -232,3 +233,74 @@ def append_reload_and_write(self, s, filename, encoding):
                 # Make sure that entries contains at least one element. If the
                 # file and s are empty, this can occur.
                 self.entries = ['']
+
+class BPythonLinecache(dict):
+    """Replaces the cache dict in the standard-library linecache module,
+    to also remember (in an unerasable way) bpython console input."""
+
+    def __init__(self, *args, **kwargs):
+        dict.__init__(self, *args, **kwargs)
```
sebastinas: `dict` is new style class, isn't it? Please use `super` instead.

bpython/history.py
```diff
@@ -232,3 +233,74 @@ def append_reload_and_write(self, s, filename, encoding):
...
+linecache.cache = BPythonLinecache(linecache.cache)
+linecache.clearcache = _bpython_clear_linecache
+
+def filename_for_console_input(code_string):
+    """Remembers a string of source code, and returns
+    a fake filename to use to retrieve it later."""
+    try:
+        return linecache.cache.remember_bpython_input(code_string)
+    except AttributeError:
+        return '<input>'
```
sebastinas: Is there a reason for catching the `AttributeError`s? We've set it to our implementation after all.

bpython/history.py
```diff
@@ -232,3 +233,74 @@ def append_reload_and_write(self, s, filename, encoding):
                 # Make sure that entries contains at least one element. If the
                 # file and s are empty, this can occur.
                 self.entries = ['']
+
+class BPythonLinecache(dict):
```
sebastinas: I think I'd prefer this in a separate file. `bpython/history.py` is for handling the history file.

rhymeswithcycle:
Sure, moved things to a separate file.

The motivation for catching `AttributeError`s is if some other code also patches the same object -- if that happens, better to just no longer have `inspect.getsource()` work for interactively-defined functions than to raise an exception.

Issue #506, opened 2015-03-11
Incorrect behavior on bpython -i module.py and bpython module.py


thomasballinger:
https://bpaste.net/show/5ae7b297bffe is code which when run with `bpython -i module.py` or `bpython module.py` doesn't match the behavior of the same options used with python. Reported by @RichardBronosky

RichardBronosky:
This is what I've found out so far:
https://github.com/bpython/bpython/blob/8aa4ddfefbb6963d325e7aa6ef06c0ab8a9f5d01/bpython/cli.py#L1907 calls https://github.com/bpython/bpython/blob/master/bpython/args.py#L118 which bubbles up to https://github.com/bpython/bpython/blob/7eaef534e05de3d06a8c076bf5737050349bba89/bpython/repl.py#L112 then https://github.com/python-git/python/blob/master/Lib/code.py#L87 then https://github.com/python-git/python/blob/master/Lib/code.py#L103

I noticed that the last line there looks just like http://code.activestate.com/recipes/82234-importing-a-dynamically-generated-module/ which has me wondering... Can we use `imp` to create a new module and set its `__dict__` equal to that `code.InteractiveInterpreter` instances `locals` var. Then set sys.modules['__console__'] to that new module.

That would at least provide access to the module that was executed via cli arg. This would still require special casing for any code that expected to find the module at sys.modules['__main__'] but it would work for sys.modules[__name__]

thomasballinger:
I like this approach, you're thinking something like this right? I wonder what the repercussions of swapping out __main__ would be.
~~~
(bpython)tom-mba:bpython tomb$ git diff bpython/args.py
diff --git a/bpython/args.py b/bpython/args.py
index 6fd2b51..91c8fc7 100644
--- a/bpython/args.py
+++ b/bpython/args.py
@@ -3,9 +3,10 @@ Module to handle command line argument parsing, for all front-ends.
 """

 from __future__ import print_function
+import code
+import imp
 import os
 import sys
-import code
 from optparse import OptionParser, OptionGroup

 from bpython import __version__
@@ -114,6 +115,9 @@ def exec_code(interpreter, args):
         source = sourcefile.read()
     old_argv, sys.argv = sys.argv, args
     sys.path.insert(0, os.path.abspath(os.path.dirname(args[0])))
+    mod = imp.new_module('__console__')
+    sys.modules['__console__'] = mod
+    interpreter.locals = mod.__dict__
     interpreter.locals['__file__'] = args[0]
[...]

PR #525, opened 2015-04-26
Fix 506

thekthuser:
Included the code Tom suggested at https://github.com/bpython/bpython/issues/506


thomasballinger:
whoops, to wrong branch

Issue #474, opened 2015-02-07
Path mangling for hacky virtualenvironment support


thomasballinger:
IPython [changes sys.path to make virtual environments kinda sorta work](https://github.com/ipython/ipython/blob/master/IPython/core/interactiveshell.py#L726). I wonder if we should do this?

sebastinas:
I don't think we should implement something like this. Too much magic, hacky and deviates from the default behaviour.

thomasballinger:
sys.path trickery seems like a bad idea

Issue #353, opened 2014-08-21
Use subprocess for interpreter


thomasballinger:
This would:
- make supporting virtual environments pretty easy
- make module reloading more reliable (no more not being able to reload some modules, no more having access to the same modules as bpython)
- get rid of the greenlet dependency (requires c compiler) - no more jumping back and forth between of user code and UI code
- make getting threads to work normally easier
- allow a cleaner evented system around a select loop
- make user subprocesses work correctly
- install one version of bpython (python2.6 - 3.4) and bpython works with all python versions
- let bipython reuse more of our code maybe

It would suck because:
- we would have to do all the completion with interprocess communication
- interprocess communication might not be fast enough for the user experience we need - and leads to caching which could get complicated in the best case, and not work in the worst
- Windows compatibility would probably go further out the window
- some code will have to run in the process (import completion, import observer, completion introspection)
- a simple python repl probably isn't enough - we'll still need an InteractiveInterpreter over there
- there are going to be difficulties with version compatibilities - every combination (n by n) of Python versions ought to be tested
- this may involve reading the IPython code and noticing that it's nicer in some ways than ours :wink:

thomasballinger:
New thoughts on this:
* if we don't use the Jupyter messaging api, this should be done with https://github.com/pexpect/pexpect
* maybe implement a fun forking thing, similar to https://github.com/thomasballinger/rlundo

PR #532, opened 2015-05-04
Load virtualenv if detected

romain-dartigues:
Hi,

this patch allow a **system-wide** installation of `bpython` to run a virtualenv; ie:

```console
user@darkstar:~$ type bpython
bpython is hashed (/usr/local/bin/bpython)
user@darkstar:~$ workon sandbox
(sandbox)user@darkstar:~$ bpython
```
```python
bpython version 0.15.dev115 on top of Python 2.7.3 /usr/bin/python
>>> import sys
>>> sys.real_prefix # see: http://stackoverflow.com/a/1883251/248390
'/usr'
>>> sys.path
['', '/home/user/.virtualenvs/sandbox/lib/python2.7/site-packages', ...]
```

Automatic loading of the virtualenv may be disabled with: `--no-virtualenv`.

I humbly think this feature is desired by the community.

### DIFF
--- a/bpython/args.py
+++ b/bpython/args.py
@@ -75,6 +75,10 @@ def parse(args, extras=None, ignore_stdin=False):
                       help=_("Don't flush the output to stdout."))
     parser.add_option('--version', '-V', action='store_true',
                       help=_('Print version and exit.'))
+    parser.add_option('--virtualenv', action='store_true', default=True,
+	    help=_('Load virtualenv if found (default: %default).'))
+    parser.add_option('--no-virtualenv', dest='virtualenv',
+	    action='store_false', help=_('Do *not* load virtualenv if found.'))
 
     if extras is not None:
         extras_group = OptionGroup(parser, extras[0], extras[1])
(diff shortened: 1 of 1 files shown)


romain-dartigues:
I would be glad if someone could help me to fix `TestExecArgs` failures.

In my understanding Travis have a VIRTUAL_ENV variable (set to "python3.3.5" or "python3.4.2" in the failing cases) which is unreachable at the time of the tests.
I'm not sure how you want this to  get fixed.

sebastinas:
I think this feature is considered part of the "too much magic" category and was explicitly NAKed some time ago. 

If we had the split between the interpreter and the interface in different process, this could be done by running the interpreter in the virtual env and the interface outside.

romain-dartigues:
Hi @sebastinas, thank you for your reply.

Sorry, I did not find a reference indicating this feature was rejected in the past.
However i insist i think it might be appreciated by some users (me me me!) to have the ""magic"" out of the box, maybe disabled by default or a shortcut inside the CLI, but at least a switch to turn it on easily (at least in a easier way than executing the snipped at hand)?

Regards.

sebastinas:
#7 and #474 from a quick search. There might be more in the old bitbucket issue tracker.

Anyway, I'd rather have proper separation of the interface and the interpreter. We'd get this feature for free. But as long as the implementation of  venv auto-detection affects the interface part and might break bpython in horrible ways (incompatible libraries in the venv), I don't think this is a good idea.

romain-dartigues:
OK, haven't found those issues in my previous search...
Do you think it is still possible to "properly separate" the interface and the interpreter?
If it is not "too late", would you be available to discuss outside this thread about how you see this separation implemented, in the event that I would have time to work on it.

sebastinas:
@romain-dartigues: If you are interested in implementing the split, we can continue the discussion in #353.

Issue #528, opened 2015-04-28
Show single underscore attributes before double underscore ones


thomasballinger:
~~~
>>> n.__dict__
┌──────────────────────────────────────────────────────────────────────────────┐
│ __class__          __delattr__        __dict__                               │
│ __doc__            __format__         __getattribute__                       │
│ __hash__           __init__           __module__                             │
│ __new__            __reduce__         __reduce_ex__                          │
│ __repr__           __setattr__        __sizeof__                             │
│ __str__            __subclasshook__   __weakref__                            │
│ _foo                                                                         │
└──────────────────────────────────────────────────────────────────────────────┘
~~~
I think it would be nice to show the single underscore methods first, then the doubles - maybe even to hold off on the double underscore suggestions until the user types a second underscore?

PR #561, opened 2015-09-10
Show single underscore attributes before double underscore ones #528

wevial:
Autocomplete shows single underscore attributes before double underscore ones, as per Tom's suggestion.

### DIFF
--- a/bpython/autocomplete.py
+++ b/bpython/autocomplete.py
@@ -256,7 +256,6 @@ class AttrCompletion(BaseCompletionType):
         if not r.word.split('.')[-1].startswith('_'):
             matches = set(match for match in matches
                           if not match.split('.')[-1].startswith('_'))
-        # MAYBE HERE 2 ??
         return matches
 
     def locate(self, current_offset, line):
@@ -312,8 +311,6 @@ class AttrCompletion(BaseCompletionType):
         for word in words:
             if self.method_match(word, n, attr) and word != "__builtins__":
                 matches.append("%s.%s" % (expr, word))
-#        print 'matches', matches
-        # ORGANIZE THE THINGS HERE MAYBE 1??
         return matches
 
     if py3:
--- a/bpython/autocomplete.py
+++ b/bpython/autocomplete.py
@@ -559,10 +559,10 @@ def sort_by_underscore(matches):
     if len(matches) == 0 or len(matches) == 1:
         return matches
     one_underscore = None
+    dot = matches[i].rfind('.') + 1
     for i in xrange(1, len(matches)+1):
         i = -i
-        dot_i = matches[i].rfind('.') + 1
-        if matches[i][dot_i] == '_' and matches[i][dot_i+1] != '_':
+        if matches[i][dot] == '_' and matches[i][dot+1] != '_':
             one_underscore = i
         else:
             break
--- a/bpython/autocomplete.py
+++ b/bpython/autocomplete.py
@@ -558,17 +558,10 @@ def sort_by_underscore(matches):
     matches = sorted(matches)
     if len(matches) == 0 or len(matches) == 1:
         return matches
-    one_underscore = None
-    dot = matches[i].rfind('.') + 1
-    for i in xrange(1, len(matches)+1):
-        i = -i
-        if matches[i][dot] == '_' and matches[i][dot+1] != '_':
-            one_underscore = i
-        else:
-            break
-    if one_underscore != None:
-        return matches[one_underscore:] + matches[:one_underscore]
-    return matches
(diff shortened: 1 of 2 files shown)


bpython/autocomplete.py
```diff
@@ -256,6 +256,7 @@ def matches(self, cursor_offset, line, **kwargs):
         if not r.word.split('.')[-1].startswith('_'):
             matches = set(match for match in matches
                           if not match.split('.')[-1].startswith('_'))
+        # MAYBE HERE 2 ??
```
thomasballinger: Looks like a few comments made it in by accident

bpython/test/test_autocomplete.py
```diff
@@ -88,9 +88,11 @@ def test_only_completer_returns_None(self):
     def test_first_completer_returns_None(self):
         a = completer(None)
         b = completer(['a'])
+        print '\n\n\n', b.matches, '\n\n\n'
```
thomasballinger: remove this

bpython/test/test_autocomplete.py
```diff
@@ -365,3 +367,6 @@ def func(apple, apricot, banana, carrot):
                             set(['banana=']))
         self.assertSetEqual(com.matches(3, "car", argspec=argspec),
                             set(['carrot=']))
+
```
thomasballinger: Instead of this, you can run individual tests with `nose tests/test_autocomplete.py`

bpython/autocomplete.py
```diff
@@ -543,13 +546,31 @@ def get_completer(completers, cursor_offset, line, **kwargs):
...
-            return sorted(matches), (completer if matches else None)
+            if len(matches) > 0:
+                matches = sort_by_underscore(matches)
+            return matches, (completer if matches else None)
     return [], None
 
+def sort_by_underscore(matches):
+    """Sort single underscore attributes before double underscore ones.
+    """
+    matches = sorted(matches)
```
thomasballinger: Doesn't really matter, but I'd prefer `matches.sort()` instead of reassigning to the variables - I guess this keeps the matches list being passed in safe from modifications, but if we're doing this we should change the docstring from "sort asdfasdf" to "Returns a sorted list" because sort implies mutating the list imo.

bpython/autocomplete.py
```diff
@@ -543,13 +546,31 @@ def get_completer(completers, cursor_offset, line, **kwargs):
...
+    if len(matches) == 0 or len(matches) == 1:
+        return matches
+    one_underscore = None
+    for i in xrange(1, len(matches)+1):
```
thomasballinger: I bet we could do this with a single `[].sort(key=lambda sug: (sug.startswith('_'), sug.startswith('__'), sug))` or similar

thomasballinger: Or return 'sorted([m for m in matches if not m.startswith('_')]) + sorted([m for m in matches if m.startswith('_')]) + sorted([m for m in matches if m.startswith('__')])`

thomasballinger:
Also check out the tests, https://travis-ci.org/bpython/bpython/jobs/79684995

bpython/autocomplete.py
wevial: The list of matches starts with the name of the variable, etc. before the attribute (eg. ['t.__class__', 't.__format__', 't.__repr__', 't.__init__', 't._bar', 't._foo'] ) so that doesn't quite work unfortunately

wevial: Unless I'm sorting in the wrong place, which is totally plausible

thomasballinger: Hm I didn't actually test these so they might not be right, but there just might be some nice way similar to one of these.

wevial: Also, ```[m for m in matches if m.startswith('_')]``` still selects the double underscore attributes.

thomasballinger: Ah ok so these won't actually work, but if there is some nice way to do
this in several steps without the for loop it could be nice. Either way,
the only real blocker right now is getting it passing the tests and getting
prints/comments out.

thomasballinger:
Looks great, thanks @wevial. I'll take a closer look today. Hope you're well!

thomasballinger:
Ah I see - this is absolutely my fault, but because I didn't review this in time https://github.com/bpython/bpython/pull/568 was merged instead. That PR doesn't quite do what I requested in issue #528, it doesn't even show double underscore methods until the user adds the second underscore. But it does the basic idea, and it's maybe a better idea.

@wevial this is great but looks like we don't need it. Thanks for submitting and doing the changes I asked for, sorry I didn't review it in time because I absolutely would have accepted it.

PR #568, opened 2015-10-01
Filter out two underscore attributes in auto completion

Tenzer:
This change will require you to write two underscores in order to get autocompletion of attributes starting with two underscores, as requested in #528.

I'm not too sure about the changes to the tests, and if there should be some tests for this filtering as well. I just tried to make the change as minimal as possible while making the tests pass.

### DIFF
--- a/bpython/test/test_autocomplete.py
+++ b/bpython/test/test_autocomplete.py
@@ -245,12 +245,12 @@ class TestAttrCompletion(unittest.TestCase):
                                              locals_={'a': OldStyleFoo()}),
                             set(['a.method', 'a.a', 'a.b']))
         self.assertIn(u'a.__dict__',
-                      self.com.matches(3, 'a._', locals_={'a': OldStyleFoo()}))
+                      self.com.matches(3, 'a.__', locals_={'a': OldStyleFoo()}))
 
     @skip_old_style
     def test_att_matches_found_on_old_style_class_object(self):
         self.assertIn(u'A.__dict__',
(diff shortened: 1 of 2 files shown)


bpython/test/test_autocomplete.py
```diff
@@ -245,12 +245,12 @@ def test_att_matches_found_on_old_style_instance(self):
                                              locals_={'a': OldStyleFoo()}),
                             set(['a.method', 'a.a', 'a.b']))
         self.assertIn(u'a.__dict__',
-                      self.com.matches(3, 'a._', locals_={'a': OldStyleFoo()}))
+                      self.com.matches(3, 'a.__', locals_={'a': OldStyleFoo()}))
```
sebastinas: If I remember correctly, the first argument should be changed to 4. @thomasballinger, can you confirm that?

Tenzer: The tests pass with the value set to 3. I also think that makes sense with it finding `a`, `__init__` and `method`. Which attribute would you say is missing?

sebastinas: `3` and `a._` means to test for `a._<Tab>` while `3` and `a.__` means to test for `a._<Tab>_` which is a different test. The first argument is the position of the cursor at the time of the autocompletion call. So leaving the first value set to `3` does not preserve the meaning of the test.

Tenzer: Oh, my bad. The it totally makes sense to change it to 4. I have done that in 9f3460b.

PR #569, opened 2015-10-07
Fix small python 3 compatibility issue using windows

ObviouslyGreen:
Something I bumped into getting bpython on windows. Python 3 doesn't allow addition of dict_items and list in make_colors if OS is windows, but union works!


thomasballinger:
Whoops... this seems to have broken Python 2 compatibility! @adityagupta679 is working on the fix.

Issue #37, opened 2014-07-04
autocomplete could work for items in lists, tuples, etc


ikanobori:
In the following case (where -!- is the cursor position):

> > > l = ['foo', 'bar', 'baz']
> > > l[0].-!-

It won't autocomplete with string methods.

This makes sense in the general case, because l[0] could be a method call, but if type(l) is list, bpython could autocomplete according with the type of l[0] without any side effects. The same could be done for tuples, and dicts.

It may be reasonably safe to extend this to subclasses of list, tuple and dicts too. This way it would work for nametuples, defaultdicts, etc...

---
- Bitbucket: https://bitbucket.org/bobf/bpython/issue/37
- Originally Reported By: [ ](http://bitbucket.org/nilton)
- Originally Created At: 2009-06-15T03:43:57.160

axs221:
FYI I started working on this tonight. 

I am working on the main case mentioned at the moment:

> l = ['foo', 'bar', 'baz']
> l[0].-!-

It is working at the moment, but only if there are string literals inside the array, and it doesn't handle out of bounds, etc. I'll get a PR before long.

thomasballinger:
Current status:

The original portion of this issue has been fixed by @axs221, though it's not side-effect proof yet. The later comment about completion on literals is done for strings, but no other types yet. Using literal_eval as suggested by @sebastinas is probably the right way to do this in a general way.

thomasballinger:
Update: after https://github.com/bpython/bpython/commit/05c83beb32b1c754e127fe40f8e09c7d2d3abf4d array completion is more limited again, though safe. I'm thinking about a literal_eval sort of thing that also allows variable lookup.

thomasballinger:
I'm taking a shot at this with https://github.com/bpython/bpython/pull/580

PR #577, opened 2015-11-27
Autocomplete for lists, tuples - autocomplete could work for items in lists, tuples, etc #37

axs221:
This handles part of #37. 

It so far works with array indexing (lists and tuples). It was tested mainly for strings and objects. It does support nesting of arrays:
a[0][0].

I will later looking into supporting dictionaries, and possibly inline creation of data structures.

### DIFF
--- a/bpython/autocomplete.py
+++ b/bpython/autocomplete.py
@@ -349,6 +349,7 @@ class ArrayObjectMembersCompletion(BaseCompletionType):
         r = self.locate(cursor_offset, line)
         if r is None:
             return None
+        member_part = r[2]
         _, _, dexpr = lineparts.current_array_with_indexer(cursor_offset, line)
         try:
             obj = safe_eval(dexpr, locals_)  # TODO
@@ -360,6 +361,7 @@ class ArrayObjectMembersCompletion(BaseCompletionType):
             obj = '"' + obj + '"'
 
         obj += "."
+        obj += member_part
 
         for completer in self.completers:
             matches = completer.matches(len(obj), obj)
--- a/bpython/autocomplete.py
+++ b/bpython/autocomplete.py
@@ -335,6 +335,44 @@ class AttrCompletion(BaseCompletionType):
                 return dir(obj)
 
 
+class ArrayObjectMembersCompletion(BaseCompletionType):
+
+    def __init__(self, shown_before_tab=True, mode=SIMPLE):
+        self._shown_before_tab = shown_before_tab
+        self.completers = [ StringLiteralAttrCompletion() ]
+
+    def matches(self, cursor_offset, line, **kwargs):
+        if 'locals_' not in kwargs:
+            return None
+        locals_ = kwargs['locals_']
+
+        r = self.locate(cursor_offset, line)
+        if r is None:
+            return None
+        _, _, dexpr = lineparts.current_array_with_indexer(cursor_offset, line)
+        try:
+            obj = safe_eval(dexpr, locals_)  # TODO
+            # obj = eval(dexpr)
+        except EvaluationError:
+            return set()
+
+        if isinstance(obj, str):
+            obj = '"' + obj + '"'
+
+        obj += "."
+
+        for completer in self.completers:
+            matches = completer.matches(len(obj), obj)
+            if matches:
+                return matches
+
+    def locate(self, current_offset, line):
+        return lineparts.current_array_item_member_name(current_offset, line)
+
+    def format(self, match):
+        return match[:-1]
+
+
 class DictKeyCompletion(BaseCompletionType):
 
     def matches(self, cursor_offset, line, **kwargs):
@@ -565,6 +603,7 @@ def get_default_completer(mode=SIMPLE):
         MagicMethodCompletion(mode=mode),
         MultilineJediCompletion(mode=mode),
         GlobalCompletion(mode=mode),
+        ArrayObjectMembersCompletion(mode=mode),
         CumulativeCompleter((AttrCompletion(mode=mode),
                              ParameterNameCompletion(mode=mode)),
                             mode=mode)
--- a/bpython/autocomplete.py
+++ b/bpython/autocomplete.py
@@ -339,7 +339,11 @@ class ArrayObjectMembersCompletion(BaseCompletionType):
 
     def __init__(self, shown_before_tab=True, mode=SIMPLE):
         self._shown_before_tab = shown_before_tab
-        self.completers = [ StringLiteralAttrCompletion() ]
+        self.completers = CumulativeCompleter((
+                            StringLiteralAttrCompletion(mode=mode),
+                            AttrCompletion(mode=mode),
+                            ParameterNameCompletion(mode=mode)),
+                            mode=mode)
 
     def matches(self, cursor_offset, line, **kwargs):
         if 'locals_' not in kwargs:
@@ -352,21 +356,16 @@ class ArrayObjectMembersCompletion(BaseCompletionType):
         member_part = r[2]
         _, _, dexpr = lineparts.current_array_with_indexer(cursor_offset, line)
         try:
-            obj = safe_eval(dexpr, locals_)  # TODO
-            # obj = eval(dexpr)
+            exec('temp_val_from_array = ' + dexpr, locals_)
         except EvaluationError:
             return set()
+        except NameError:
+            return set()
 
-        if isinstance(obj, str):
-            obj = '"' + obj + '"'
+        obj = 'temp_val_from_array.'
 
-        obj += "."
-        obj += member_part
-
-        for completer in self.completers:
-            matches = completer.matches(len(obj), obj)
-            if matches:
-                return matches
+        matches = self.completers.matches(len(obj), obj, **kwargs)
+        return [match.replace('temp_val_from_array.', line) for match in matches]
 
     def locate(self, current_offset, line):
         return lineparts.current_array_item_member_name(current_offset, line)
--- a/bpython/autocomplete.py
+++ b/bpython/autocomplete.py
@@ -374,7 +374,7 @@ class ArrayItemMembersCompletion(BaseCompletionType):
         return lineparts.current_array_item_member_name(current_offset, line)
 
     def format(self, match):
-        return match[:-1]
+        return after_last_dot(match)
 
 
 class DictKeyCompletion(BaseCompletionType):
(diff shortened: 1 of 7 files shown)


thomasballinger:
This is great, thanks for working on this!

I haven't looked at the code much, one though based on the functionality: right now it seems like the wrong part of each suggestion is being displayed:

    (bpython)tomb@tom-mba ((2472bb0...)) bpython$ bpython
    bpython version 0.15.dev179 on top of Python 2.7.10 /Users/tomb/.virtualenvs/bpython/bin/python2.7
    >>> l = [1,2,3,4]
    >>> l[0].
    ┌──────────────────────────────────────────────────────────────────────────────────────┐
    │ l[0].bit_lengt   l[0].conjugat    l[0].denominato  l[0].ima                          │
    │ l[0].numerato    l[0].rea                                                            │
    └──────────────────────────────────────────────────────────────────────────────────────┘

I think preferred behavior would be stripping out the `l[0].` portion, which seems to be how width of each suggestion is being calculated already.

axs221:
Hmm I can take a look at that, though note I actually have different
results than you, it isn't truncated for me, e.g. "bit_length" is spelled
out:

>>> l = [1,2,3,4]
>>> l[0].
┌────────────────────────────────────────────────────────────────────────────┐
│l[0].bit_length  l[0].conjugate   l[0].denominator l[0].imag
 │
│l[0].numerator   l[0].real
 │
└────────────────────────────────────────────────────────────────────────────


Also, I was basing this off of AttrCompletion, which also shows the name. I
am happy to change it, but we could consider changing others as well for
consistency:


>>> l = dict()
>>> l.
┌────────────────────────────────────────────────────────────────────────────┐
│l.clear      l.copy       l.fromkeys   l.get        l.has_key
  │
│l.items      l.iteritems  l.iterkeys   l.itervalues l.keys
 │
│l.pop        l.popitem    l.setdefault l.update     l.values
 │
│l.viewitems  l.viewkeys   l.viewvalues
 │
└────────────────────────────────────────────────────────────────────────────






Shawn Axsom

thomasballinger:
Whoa weird, that's not the behavior I see. Hmm, looking into this. Locally I get only the attributes:

    (bpython)tomb@tom-mba ~$ bpython                   │(bpython3)tomb@tom-mba ~$ bpython
    bpython version 0.15.dev179 on top of Python 2.7.10│bpython version 0.15.dev179 on top of Python 3.5.0
     /Users/tomb/.virtualenvs/bpython/bin/python2.7    │ /Users/tomb/.virtualenvs/bpython3/bin/python3
    >>> l = dict()                                     │>>> l = dict()
    >>> l.                                             │>>> l.
    ┌──────────────────────────────────────┐           │┌──────────────────────────────────────┐
    │ clear        copy                    │           ││ clear        copy                    │
    │ fromkeys     get                     │           ││ fromkeys     get                     │
    │ has_key      items                   │           ││ items        keys                    │
    │ iteritems    iterkeys                │           ││ pop          popitem                 │
    │ itervalues   keys                    │           ││ setdefault   update                  │
    │ pop          popitem                 │           ││ values                               │
    │ setdefault   update                  │           │└──────────────────────────────────────┘
    │ values       viewitems               │           │
    │ viewkeys     viewvalues              │           │
    └──────────────────────────────────────┘           │

thomasballinger:
Yeah locally with your PR I get behavior that matches the last released version, which is what currently runs at http://ballingt.com/bpython-demo/

thomasballinger:
Ah maybe you're using bpython-urwid? If so, that makes sense and your code looks fine there. I can take a look at getting the completer to work with both as expected if you like.

Separately if you or another bpython-urwid user prefers this just-attributes behavior another issue would be welcome, I don't use bpython-urwid much so don't have strong feelings about it.

thomasballinger:
for the record, here's what the urwid frontend looks like for me:

    >>> l = []                                         │>>> l = []
    >>> l = [1,2,3]                                    │>>> l = [1,2,3]
    >>> l[0].                                          │>>> l[0].
    ┌─────────────────────────────────────────────────┐│┌────────────────────────────────────────────────┐
    │l[0].bit_length  l[0].conjugate                  │││l[0].bit_length  l[0].conjugate                 │
    │l[0].denominator l[0].imag                       │││l[0].denominator l[0].from_bytes                │
    │l[0].numerator   l[0].real                       │││l[0].imag        l[0].numerator                 │
    └─────────────────────────────────────────────────┘││l[0].real        l[0].to_bytes                  │
                                                       │└────────────────────────────────────────────────┘

axs221:
Hmm is my PR incorrect then? I get the same behavior as my PR if I clone
master:

   - git clone https://github.com/bpython/bpython.git

Do you know how I can get the source version where AttrCompletion doesn't
show the variable name?



Shawn Axsom

thomasballinger:
My only guess is that you're using bpython-urwid (maybe via an alias?) instead of bin/bpython

axs221:
Ah gotcha, yeah that is the case, yeah I was running the urwid version from
a script. I'll look into it along with a few unit test issues. Thanks!


Shawn Axsom

axs221:
I just pushed a change to the PR that should fix the display, I'll look at
any remaining unit test issues next.


Shawn Axsom

axs221:
Checks passed, let me know if you need anything else with this PR, otherwise I'll assume it will be merged and I will look into other features (esp. dictionaries) regarding #577 soon!

bpython/autocomplete.py
```diff
@@ -335,6 +335,48 @@ def list_attributes(self, obj):
                 return dir(obj)
 
 
+class ArrayItemMembersCompletion(BaseCompletionType):
+
+    def __init__(self, shown_before_tab=True, mode=SIMPLE):
+        self._shown_before_tab = shown_before_tab
+        self.completers = CumulativeCompleter((
+                            StringLiteralAttrCompletion(mode=mode),
```
thomasballinger: I can't figure out what would cause this to trigger, aren't we always completing an attribute?

bpython/autocomplete.py
```diff
@@ -335,6 +335,48 @@ def list_attributes(self, obj):
...
+            return None
+        locals_ = kwargs['locals_']
+
+        r = self.locate(cursor_offset, line)
+        if r is None:
+            return None
+        member_part = r[2]
+        _, _, dexpr = lineparts.current_array_with_indexer(cursor_offset, line)
+        try:
+            exec('temp_val_from_array = ' + dexpr, locals_)
```
thomasballinger: Is there a reason to do it this way instead of safe_eval like other completers?

bpython/autocomplete.py
axs221: Good catch. Yeah I don't think that is necessary any more. I was using it in a previous version of this code, but no longer need it for the "temp_val_from_array" logic I am now using. I'll remove all but AttrCompletion.

bpython/autocomplete.py
```diff
@@ -335,6 +335,48 @@ def list_attributes(self, obj):
...
+        except (EvaluationError, NameError, IndexError, AttributeError, TypeError):
+            return set()
+
+        temp_line = line.replace(member_part, 'temp_val_from_array.')
+
+        matches = self.completers.matches(len(temp_line), temp_line, **kwargs)
+        matches_with_correct_name = \
+            set([match.replace('temp_val_from_array.', member_part) for match in matches])
+
+        exec('temp_val_from_array = None', locals_)
```
thomasballinger: If this is stuck in locals, might be better to delete the binding than to set it to None

bpython/autocomplete.py
axs221: I wasn't sure how to do it with safe_eval. From my understanding, safe_eval doesn't allow me to set a temp variable dynamically, it will just return a real variable value. I needed a variable name in locals_ that I could reference by string in completers like AttrCompletion, so I could make use of their logic without any changes to their code.

thomasballinger: Ah got it, that makes sense. Now that it's just AttrCompletion, maybe using
AttrCompletion.attr_matches would work, or just duplicating that bit of
logic.

thomasballinger: Also I'm happy to make these changes, I've got it open now at can add them
to the merge commit.

On Sat, Nov 28, 2015 at 1:22 PM, Thomas Ballinger <thomasballinger@gmail.com

axs221: I went ahead and submitted some updates. I got it using safe_eval if you're fine with that approach. I just had to set locals_ acting as a dictionary, equal to the safe_eval.

thomasballinger:
This makes sense from a code reuse standpoint, though it feels like gymnastics to create a fake line and then use AttrCompletion. I was trying various things I didn't like, so this seems good. I'll merge it and then  think about refactoring it since I'm not sure I'll come up with anything.

axs221:
I agree, I didn't like creating a temp variable, though I don't know of a
solution as flexible and simple. Another idea was to break AttrCompleter
match into two, extracting a new method that could accept a regex match
possibly. But that seemed less flexible and touched existing code more than
I wanted to.

thomasballinger:
:+1:

thomasballinger:
This has been merged, but I rebased it so it's not recognizing it. Thanks so much @axs221!

axs221:
Glad to help! I will work on Dictionaries soon.

thomasballinger:
Cool! Dictionaries that are values should be working already:

    tomb@tom-mba ~$ bpython
    bpython version 0.14.2 on top of Python 2.7.10 /usr/local/opt/python/bin/python2.7
    >>> d = {123:234, 234:345, 125555:123}
    >>> d[1
    ┌─────────────────────────────────────────────────────────────┐
    │ 123     125555                                              │
    └─────────────────────────────────────────────────────────────┘

However dictionary *literals* don't have method completion, ie `{1:2}.c` doesn't bring up `.clear` and `.copy`.

axs221:
Yep that is what I was thinking, e.g.
>>> f = [1,2,3,4]
>>> f = { "a", "b" }
>>> f["a"].

thomasballinger:
Oh sorry I was confused - yeah that'd be great! I'm looking at making this safer atm, eg 

    def test_other_getitem_methods_not_called(self):
        class FakeList(object):
            def __getitem__(inner_self, i):
                self.fail("possibly side-effecting __getitem_ method called")

        self.com.matches(5, 'a[0].', locals_={'a': FakeList()})

this might carry over to how dictionaries completion would work - for both, we'll need code to evaluate, bit by bit, each time checking to see if we anticipate the object (hopefully a dictionary, list, tuple etc.) possibly having side effects on lookup.

axs221:
Sounds good to me. Let me know if you come up with anything, I will hold
off on going too far on the dictionary implementation until then.


Shawn Axsom

PR #579, opened 2015-11-29
fix parameter name completion

thomasballinger:
I found parameter name completion wasn't working. I don't know if it's ever worked. Now this happens:

    bpython version 0.15.dev177 on top of Python 2.7.10 /Users/tomb/.virtualenvs/bpython/bin/python2.7
    >>> def foo(abc=1, abd=2, xyz=3):
    ...     "function with keyword parameters"
    ...
    >>> foo(ab
    ┌──────────────────────────────────────────────────┐
    │ foo: (abc=1, abd=2, xyz=3)                       │
    │ abc= abd= abs(                                   │
    │ function with keyword parameters                 │
    └──────────────────────────────────────────────────┘

### DIFF
--- a/bpython/autocomplete.py
+++ b/bpython/autocomplete.py
@@ -160,17 +160,17 @@ class CumulativeCompleter(BaseCompletionType):
         return self._completers[0].format(word)
 
     def matches(self, cursor_offset, line, **kwargs):
+        return_value = None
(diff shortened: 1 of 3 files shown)


thomasballinger:
I've played with this a bit, it doesn't appear to break anything.

PR #580, opened 2015-11-30
Attribute completion on simple expressions

thomasballinger:
This implements the feature initially described in issue #37, autocompletion of literals, and includes autocompletion of literals as discussed later in the thread.

This implementation is relatively conservative: it will only evaluate lists, tuples, strs, unicodes, dicts. It evaluates no methods called with method call syntax, only `__getitem__` invoked by `[]` on these types.

To find an expression to evaluate before current attribute being completed, ast.parse is called on strings of increasing size - this could be a performance issue.

### DIFF
--- a/bpython/simpleeval.py
+++ b/bpython/simpleeval.py
@@ -8,6 +8,8 @@ In order to provide fancy completion, some code can be executed safely.
 from ast import *
 from six import string_types
 
+from bpython import line as line_properties
+
 class EvaluationError(Exception):
     """Raised if an exception occurred in safe_eval."""
 
@@ -21,7 +23,7 @@ def safe_eval(expr, namespace):
         # raise
         raise EvaluationError
 
-def simple_eval(node_or_string, namespace={}):
+def simple_eval(node_or_string, namespace=None):
     """
     Safely evaluate an expression node or a string containing a Python
     expression.  The string or node provided may only consist of:
@@ -35,6 +37,8 @@ def simple_eval(node_or_string, namespace={}):
 
     The optional namespace dict-like ought not to cause side effects on lookup
     """
+    if namespace is None:
+        namespace = {}
     if isinstance(node_or_string, string_types):
         node_or_string = parse(node_or_string, mode='eval')
     if isinstance(node_or_string, Expression):
@@ -77,37 +81,65 @@ def simple_eval(node_or_string, namespace={}):
         raise ValueError('malformed string')
     return _convert(node_or_string)
 
+
 def safe_getitem(obj, index):
     if type(obj) in (list, tuple, dict, bytes) + string_types:
         return obj[index]
     raise ValueError('unsafe to lookup on object of type %s' % (type(obj), ))
 
 
-class AttributeSearcher(NodeVisitor):
-    """Search for a Load of an Attribute at col_offset"""
-    def visit_attribute(self, node):
-        print node.attribute
+def find_attribute_with_name(node, name):
+    """Based on ast.NodeVisitor"""
+    if isinstance(node, Attribute) and node.attr == name:
+        return node
+    for field, value in iter_fields(node):
+        if isinstance(value, list):
+            for item in value:
+               if isinstance(item, AST):
+                    r = find_attribute_with_name(item, name)
+                    if r:
+                        return r
+        elif isinstance(value, AST):
+            r = find_attribute_with_name(value, name)
+            if r:
+                return r
 
-def _current_simple_expression(cursor_offset, line):
+
+def evaluate_current_expression(cursor_offset, line, namespace={}):
     """
-    Returns the current "simple expression" being attribute accessed
+    Return evaluted expression to the right of the dot of current attribute.
 
     build asts from with increasing numbers of characters.
     Find the biggest valid ast.
     Once our attribute access is a subtree, stop
-
-
     """
 
     # in case attribute is blank, e.g. foo.| -> foo.xxx|
     temp_line = line[:cursor_offset] + 'xxx' + line[cursor_offset:]
     temp_cursor = cursor_offset + 3
+    temp_attribute = line_properties.current_expression_attribute(
+            temp_cursor, temp_line)
+    if temp_attribute is None:
+        raise EvaluationError("No current attribute")
+    attr_before_cursor = temp_line[temp_attribute.start:temp_cursor]
 
-    for i in range(temp_cursor-1, -1, -1):
-        try:
-            tree = parse(temp_line[i:temp_cursor])
-        except SyntaxError:
-            return None
-        
+    def parse_trees(cursor_offset, line):
+        for i in range(cursor_offset-1, -1, -1):
+            try:
+                ast = parse(line[i:cursor_offset])
+                yield ast
+            except SyntaxError:
+                continue
 
+    largest_ast = None
+    for tree in parse_trees(temp_cursor, temp_line):
+        attribute_access = find_attribute_with_name(tree, attr_before_cursor)
+        if attribute_access:
+            largest_ast = attribute_access.value
 
+    if largest_ast is None:
+        raise EvaluationError("Corresponding ASTs to right of cursor are invalid")
+    try:
+        return simple_eval(largest_ast, namespace)
+    except (ValueError, KeyError, IndexError):
+        raise EvaluationError("Could not safely evaluate")
--- a/bpython/simpleeval.py
+++ b/bpython/simpleeval.py
@@ -168,10 +168,12 @@ def evaluate_current_expression(cursor_offset, line, namespace=None):
     """
     Return evaluated expression to the right of the dot of current attribute.
 
-    build asts from with increasing numbers of characters.
-    Find the biggest valid ast.
-    Once our attribute access is a subtree, stop
+    Only evaluates builtin objects, and do any attribute lookup.
     """
+    # Builds asts from with increasing numbers of characters back from cursor.
+    # Find the biggest valid ast.
+    # Once our attribute access is found, return its .value subtree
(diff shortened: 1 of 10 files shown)


bpython/simpleeval.py
```diff
@@ -0,0 +1,145 @@
+# encoding: utf-8
+"""simple evaluation of side-effect free code
+
+In order to provide fancy completion, some code can be executed safely.
+
+"""
+
+from ast import *
```
sebastinas: Please don't. See PEP-8 about wildcard imports.

bpython/simpleeval.py
```diff
@@ -0,0 +1,145 @@
...
+    """
+    Safely evaluate an expression node or a string containing a Python
+    expression.  The string or node provided may only consist of:
+    * the following Python literal structures: strings, numbers, tuples,
+        lists, dicts, booleans, and None.
+    * variable names causing lookups in the passed in namespace or builtins
+    * getitem calls using the [] syntax on objects of the types above
+    * getitem calls on subclasses of the above types if they do not override
+        the __getitem__ method and do not override __getattr__ or __getattribute__
+        (or maybe we'll try to clean those up?)
```
thomasballinger: This last bullet point is wrong, this isn't implemented

thomasballinger: also should mention this is based on ast.literal_eval

bpython/test/test_line_properties.py
```diff
@@ -7,7 +7,8 @@
     current_method_definition_name, current_single_word, \
     current_string_literal_attr, current_indexed_member_access_identifier, \
     current_indexed_member_access_identifier_with_index, \
```
thomasballinger: current_indexed_member_access_identifier, current_indexed_member_access_identifier_with_index, and current_indexed_member_access_member are all no longer used and should be deleted.

bpython/test/test_autocomplete.py
```diff
@@ -268,20 +268,20 @@ def __getattr__(self, attr):
...
+class TestExpressionAttributeCompletion(unittest.TestCase):
     @classmethod
     def setUpClass(cls):
-        cls.com = autocomplete.ArrayItemMembersCompletion()
+        cls.com = autocomplete.ExpressionAttributeCompletion()
 
     def test_att_matches_found_on_instance(self):
         self.assertSetEqual(self.com.matches(5, 'a[0].', locals_={'a': [Foo()]}),
-                            set(['a[0].method', 'a[0].a', 'a[0].b']))
+                            set(['method', 'a', 'b']))
```
sebastinas: Why is this changed?

thomasballinger: It's changed because this implementation (unlike TestArrayItemCompletion) never calculates what portion of the line the expression corresponds to. This should be alright because `Completer.matches()` is supposed to replace the portion of the line that its completer's `locate()` returns. ExpressionAttributeCompletion's `locate()` method returns only the attribute, like DictKeyCompletion.

bpython/test/test_line_properties.py
```diff
@@ -224,6 +225,7 @@ def test_simple(self):
         self.assertAccess('stuff[asdf[asd|fg]')
         self.assertAccess('Object.attr1.<|attr2>')
         self.assertAccess('Object.<attr1|>.attr2')
+        self.assertAccess('Object.<attr1|>.attr2')
```
thomasballinger: duplocated line

bpython/simpleeval.py
```diff
@@ -0,0 +1,145 @@
...
+        elif isinstance(node, List):
+            return list(map(_convert, node.elts))
+        elif isinstance(node, Dict):
+            return dict((_convert(k), _convert(v)) for k, v
+                        in zip(node.keys, node.values))
+        elif isinstance(node, Name):
+            try:
+                return namespace[node.id]
+            except KeyError:
+                return __builtins__[node.id]
```
thomasballinger: This is an implementation detail that breaks in pypy, import builtins instead

thomasballinger:
`class StringLiteralAttrCompletion(BaseCompletionType)` should also be removed and its test cases converted as its just a special case of this completion.

bpython/simpleeval.py
```diff
@@ -0,0 +1,145 @@
...
+        elif isinstance(node, BinOp) and \
+             isinstance(node.op, (Add, Sub)) and \
+             isinstance(node.right, Num) and \
+             isinstance(node.right.n, complex) and \
+             isinstance(node.left, Num) and \
+             isinstance(node.left.n, (int, long, float)):
```
sebastinas: What the purpose of this block?

thomasballinger: Complex numbers literals. I should have noted that this is mostly the same as the stdlib ast.literal_eval

sebastinas: Then please add a comment. The current combination of types here is confusing. Also, what's the expected result for `d[1+1j].` and `d[1j].` for `d = {1+1j: "string", 1j: "string"}`? For `d[1j].` I get member completion, while I don't get it for `d[1+1j]`.

Note that according to 2.4.3 of the Python 3 documentation there are no complex number literals (only integer, float and imaginary literals).

thomasballinger: Hm, working in Python 2 but not 3, so looks like you've hit exactly on the problem. I'll look at how much code can be shared between the two, they might have to be separate implementations.

bpython/simpleeval.py
thomasballinger: Add tests for exactly these capabilities - I didn't realize complex numbers broke in Python 3 because there weren't tests for it.

sebastinas:
Since `bpython/simpleeval.py` contains plenty of code from Python's `ast` module, the file needs proper attribution and a copyright header.

thomasballinger:
Can I use the license header from bpython/autocomplete for this? Should it instead reference to the Python License?

thomasballinger:
Thoughts on whether it's preferable to keep the simple_eval function single-source and drop the complex number literals, or make two versions that more closely mirror the corresponding ast.literal_eval implementations? This function might get more complicated in the future if the evaluation becomes less conservative (e.g. allow some attribute lookup on namedtuples).

thomasballinger:
Todo: when methods are completed they don't have function docstrings and parameters, probably because current function code can't get the function object.

thomasballinger:
@sebastinas I've added the Python license. I'd be happy to throw this code out and start again if we don't want this code in here.

Otherwise, this still needs more tests.

thomasballinger:
Also, should I add (c) bpython authors and then make an exception for those two functions?

thomasballinger:
Todo
* test that methods completed from expressions show docstrings, with the cursor at various places in the 
* factor out word.startswith('__') behavior
* test that custom number types won't have method run in simple_eval
* test that unary and binary operations work on numbers in simple_eval
* improve evaluate_current_expression docstring

bpython/simpleeval.py
```diff
@@ -0,0 +1,234 @@
...
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
```
sebastinas: Thanks.

bpython/test/test_autocomplete.py
sebastinas: Okay.

bpython/simpleeval.py
```diff
@@ -0,0 +1,234 @@
...
+        node_or_string = node_or_string.body
+
+    string_type_nodes = (ast.Str, ast.Bytes) if py3 else (ast.Str,)
+    numeric_types = (int, float, complex) + (() if py3 else (long,))
+
+    # added in Python 3.4
+    if hasattr(ast, 'NameConstant'):
+        name_type_nodes = (ast.Name, ast.NameConstant)
+    else:
+        name_type_nodes = (ast.Name,)
```
sebastinas: `string_type_nodes`, `numeric_types`, `name_type_nodes` all look like the could be global.

bpython/simpleeval.py
```diff
@@ -0,0 +1,234 @@
...
+        if isinstance(value, list):
+            for item in value:
+               if isinstance(item, ast.AST):
+                    r = find_attribute_with_name(item, name)
+                    if r:
+                        return r
+        elif isinstance(value, ast.AST):
+            r = find_attribute_with_name(value, name)
+            if r:
+                return r
```
sebastinas:
```
def find_attribute_with_name(node, name):
    if isinstance(node, ast.Attribute) and note.attr == nane:
        return node
    for item in ast.iter_child_nodes(node):
        r = find_attribute_with_name(item)
        if r:
            return r
```
should do the same and is not a code copy.

thomasballinger:
Fixed, I'm tempted to do the same thing with the other function. I've learned my lesson, I won't copy code unnecessarily in the future.

sebastinas:
There seem to be conflicts. Could you rebase the branch on top of the current master?

PR #582, opened 2015-12-03
fix run order-dependant test failure

thomasballinger:
Possibly we should have a test that explicitly tests for this behavior, but with our current tests I got failures only when tests were run in a specific order.


bpython/test/test_curtsies_repl.py
```diff
@@ -291,6 +297,7 @@ def setUp(self):
         self.open = partial(io.open, mode='wt', encoding='utf-8')
         self.dont_write_bytecode = sys.dont_write_bytecode
         sys.dont_write_bytecode = True
+        invalidate_caches()
```
sebastinas: Could you please document why `invalidate_caches` is called here?

PR #584, opened 2015-12-06
Paste detection in bpython instead of Curtsies, fixes #575

### LINKED ISSUE #575: Move paste detection into bpython instead of Curtsies
Currently a PasteEvent is returned from Curtsies when it reads more than 20 bytes in a single `fread(0)` call, but more robust paste detection would come from asking for keypress events and, if they come too fast, skipping redraws. By skipping redraws as necessary (when 2 keypress events come in before the current redraw is done) bpython could stay responsive regardless of paste events or fast typing.

Curtsies won't lose paste events for a while, but we don't have to wait for Curtsies to lose paste events to make this change.

thomasballinger:
This switches from using curtsies-created paste events to creating paste events in bpython whenever 10  or more keypress events are queued up.

I think long term this is a good move because it's one less Curtsies-specific piece of input functionality, and I'd like to remove this feature from Curtsies and use Blessed for input code eventually because I trust @jquast to get it right more than I do myself. It's also better because now it helps the UI catch up any time it gets behind on events, e.g. from a fast typist on a slow machine, whereas before the bytes had to be read in the same system call to be considered a paste event. The paste threshold of 10 should be tuned for this, it should probably be much lower (2 or 3?).

If you don't like the `combined_events` coroutine I'd be happy to reimplement it as a class, I think that might be clearer but because the `.send` and `next()` interface was all we needed it seemed easy to write that way.

### DIFF
--- a/bpython/curtsies.py
+++ b/bpython/curtsies.py
@@ -1,6 +1,7 @@
 from __future__ import absolute_import
 
 import code
+import collections
 import io
 import logging
 import sys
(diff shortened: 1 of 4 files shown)


sebastinas:
Sure

thomasballinger:
Great I'll merge it then

PR #585, opened 2015-12-06
fix #466 - don't fill whole screen with completions

### LINKED ISSUE #466: max completions box height
So that a full suggestions box doesn't move the current line to the very top of the screen, there should be a max height to the suggestions box. Could be a config value or a default if we can agree on one.

thomasballinger:
Thoughts?

* tune minimum info box size when some content is hidden (currently 30 lines)
* decide whether premptively making some room is worthwhile (currently 6 lines)
* test on small terminal sizes, make sure this doesn't make things *less* usable

I'm also going to look at making the infobox suggestions scroll.

### DIFF
--- a/bpython/curtsies.py
+++ b/bpython/curtsies.py
@@ -180,7 +180,7 @@ def mainloop(config, locals_, banner, interp=None, paste=None,
                       get_top_usable_line=get_top_usable_line) as repl:
                 repl.height, repl.width = window.t.height, window.t.width
 
-                repl.request_paint_to_pad_bottom = 10
+                repl.request_paint_to_pad_bottom = 6
 
                 def process_event(e):
                     """If None is passed in, just paint the screen"""
--- a/bpython/curtsiesfrontend/repl.py
+++ b/bpython/curtsiesfrontend/repl.py
@@ -1314,7 +1314,7 @@ class Repl(BpythonRepl):
             if self.config.curtsies_list_above:
                 info_max_rows = max(visible_space_above, visible_space_below)
             else:
-                minimum_possible_height = 4
+                minimum_possible_height = 30  # smallest an over-full completion box
                 info_max_rows = max(visible_space_below, minimum_possible_height)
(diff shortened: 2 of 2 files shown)


bpython/curtsies.py
```diff
@@ -122,6 +122,9 @@ def after_suspend():
                 window.__enter__()
                 interrupting_refresh()
 
+            def get_top_usable_line():
+                return window.top_usable_row
+
```
sebastinas: Any reason all those functions do not have a default implementation in `Repl` and are overwritten in a derived class taking  `window` and `input_generator`?

thomasballinger: As you point out, currently the Repl class doesn't have a handle on the Window object, all it has are callbacks for getting the information it needs.

You're suggesting a third class that inherits from the `curtsiesfrontend.repl.Repl` (which in turn inherits from the `repl.Repl`) that on instantiation is passed a window object and an input_generator (oh, or creating them itself would be better)? That sounds nice, I'm not sure why I did the callback style instead -- probably too much node.js and other less OOP programming recently. I don't see a reason not to, in fact the base curtsies repl class could even be functional like the old embedded version, and would be nice for testing.

thomasballinger: I'd be up for trying this refactor in a new branch, I'll open a PR early so it's easy to talk about it.

thomasballinger: Thinking back the reason was probably the frustration of hopping back and forth between bpython.curtsiesfrontent.repl.py and bpython.repl.py and thinking that for poorly defined interfaces like this (dozens of methods on each, not really specced out what belonged where) that composition was easier to keep track of.

thomasballinger: Is this is the sort of thing you're thinking off? https://github.com/bpython/bpython/compare/inheritance-over-composition

sebastinas: Yes, something like that.

PR #586, opened 2015-12-06
paginate autocompletion suggestions

thomasballinger:
This would make https://github.com/bpython/bpython/pull/585 work better, but should be useful regardless.


bpython/curtsiesfrontend/replpainter.py
```diff
@@ -48,6 +48,13 @@ def paint_current_line(rows, columns, current_display_line):
     return fsarray(lines, width=columns)
 
 
+def paginate(rows, matches, current, words_wide):
+    if current not in matches:
+        current = matches[0]
+    per_page = rows * words_wide
+    current_page = matches.index(current) // per_page
+    return matches[per_page * current_page:per_page * (current_page + 1)]
+
```
sebastinas: Please add another newline for PEP 8. Looks good otherwise.

PR #588, opened 2015-12-11
WIP Safe Callable attr parens

thomasballinger:
Adds machinery for safely doing attribute lookup (e.g. avoiding calling user-defined properties), starts completing attributes with parens if the objects to which they refer are callable, and adds attribute completion to simple expression evaluation (if #580 is merged).

One of the goals of this PR is to *safely* show completions in situations like

    >>> import ast
    >>> m = ast.parse('1 + 1')
    >>> m.body[0].value.

This isn't ready and won't be for a while. TODO:
* consolidate existing inspect.AttrCleaner functionality
* rename existing attribute lookup logic to "unsafe" or similar so its use is explicit
* test that all attribute-style completion still works with the added opening paren
* if https://github.com/bpython/bpython/pull/580 has been merged, add safe attribute completion to this
* confirm no functionality is being lost -- don't stop doing unsafe completion for attribute because this would be changing existing behavior (maybe open a new issue for this), e.g.
~~~
bpython version 0.14.2 on top of Python 2.7.10 /usr/local/opt/python/bin/python2.7
>>> class Foo():
...     @property
...     def foo(self):
...         print 1
...         return 2
...
>>> f = Foo()
1
>>> f.foo.
┌────────────────────────────────────────────────────────────────────┐
│ bit_length        conjugate         denominator                    │
│ imag              numerator         real                           │
└────────────────────────────────────────────────────────────────────┘
~~~

### DIFF
--- a/bpython/inspection.py
+++ b/bpython/inspection.py
@@ -282,38 +282,48 @@ def is_callable(obj):
     return callable(obj)
 
 
-class AttributeIsEmptySlot(object):
-    pass
-
-
 def safe_get_attribute(obj, attr):
-    """Gets attributes without triggering descriptors on new-style clases
-
-    Returns AttributeIsEmptySlot if requested attribute does not have a value,
-    but is a slot entry.
-    """
+    """Gets attributes without triggering descriptors on new-style clases"""
     if is_new_style(obj):
-        return safe_get_attribute_new_style(obj, attr)
+        result = safe_get_attribute_new_style(obj, attr)
+        if isinstance(result, member_descriptor):
+            # will either be the same slot descriptor or the value
+            return getattr(obj, attr)
+        return result
     return getattr(obj, attr)
 
 
+class _ClassWithSlots(object):
+    __slots__ = ['a']
+member_descriptor = type(_ClassWithSlots.a)
+
+
 def safe_get_attribute_new_style(obj, attr):
+    """Returns approximately the attribute returned by getattr(obj, attr)
+
+    The object returned ought to be callable if getattr(obj, attr) was.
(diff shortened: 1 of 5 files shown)


bpython/inspection.py
```diff
@@ -98,6 +98,14 @@ def __exit__(self, exc_type, exc_val, exc_tb):
         # /Dark magic
 
 
+if py3:
+    is_new_style = lambda obj: True
+else:
+    def is_new_style(obj):
+        """Returns True if obj is a new-style class or object"""
+        return type(obj) != types.InstanceType and hasattr(obj, '__class__')
```
sebastinas: What makes the  `hasattr(obj, '__class__)` necessary?

bpython/inspection.py
```diff
@@ -274,6 +282,59 @@ def is_callable(obj):
...
+def safe_get_attribute_new_style(obj, attr):
+    """Returns approximately the attribute returned by getattr(obj, attr)
+
+    The object returned ought to be callable if getattr(obj, attr) was.
+    Fake callable objects may be returned instead, in order to avoid executing
+    arbitrary code in descriptors.
+
+    If the object is an instance of a class that uses __slots__, will return
+    the member_descriptor object instead of the value.
+    """
```
sebastinas: The code doesn't seem to match the documentation.

thomasballinger:
Hm, yeah this really isn't ready. I'll close this for now.

bpython/inspection.py
thomasballinger: I can't recall, I'll take this out unless I have a test for why.

PR #589, opened 2015-12-14
Inheritance over composition

thomasballinger:
I wasn't sure how I felt about this before, but inheriting and overriding methods seems nicer than passing in a bunch of callbacks to the constructor of the Repl object.

### DIFF
--- a/bpython/curtsiesfrontend/repl.py
+++ b/bpython/curtsiesfrontend/repl.py
@@ -11,7 +11,6 @@ import signal
 import subprocess
 import sys
 import tempfile
-import threading
 import time
 import unicodedata
 from six.moves import range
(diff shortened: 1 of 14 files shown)


bpython/curtsiesfrontend/repl.py
```diff
@@ -460,7 +414,86 @@ def smarter_request_reload(files_modified=()):
...
+
+    def request_refresh(self):
+        """Arrange for the bpython display to be refreshed soon."""
+        if self.reevaluating or self.paste_mode:
+            self.fake_refresh_requested = True
+        else:
+            self._request_refresh()
+
+    def request_reload(self, files_modified=()):
+        """Arrange for the """
```
sebastinas: There's something missing.

bpython/simplerepl.py
```diff
@@ -0,0 +1,105 @@
+from __future__ import unicode_literals, print_function, absolute_import
```
sebastinas: Please add the file header with the license.

PR #592, opened 2015-12-16
fix #591 changing stack trace linenos

### LINKED ISSUE #591: normalize stack traces
https://github.com/bpython/bpython/issues/229 had the inadvertent effect of making terminal history unnecessarily inconsistent because each time a rewind happens, the input lines change:

    (bpython)tomb@tom-mba (master) bpython$ bpython
    bpython version 0.15.dev233 on top of Python 2.7.10 /Users/tomb/.virtualenvs/bpython/bin/python2.7
    >>> import
      File "<bpython-input-2>", line 1
        import
             ^
    SyntaxError: invalid syntax
    >>> 1
    1
    >>>

becomes

    (bpython)tomb@tom-mba (master) bpython$ bpython
    bpython version 0.15.dev233 on top of Python 2.7.10 /Users/tomb/.virtualenvs/bpython/bin/python2.7
    >>> import
      File "<bpython-input-4>", line 1
        import
             ^
    SyntaxError: invalid syntax
    >>>

It'd be nice to do one of the following, or something else I haven't thought of
* line count should be reset on rewind
* stack traces should be modified to remove these numbers
* history inconsistency detection should ignore these differences

thomasballinger:
Mimicking the default Python interactive interpreter, change stack traces so they no longer list the line number.

### DIFF
--- a/bpython/repl.py
+++ b/bpython/repl.py
@@ -29,6 +29,7 @@ import io
 import os
 import pkgutil
 import pydoc
+import re
 import shlex
 import subprocess
 import sys
@@ -133,6 +134,9 @@ class Interpreter(code.InteractiveInterpreter):
                 # Stuff in the right filename and right lineno
                 if not py3:
                     lineno -= 1
+                if re.match(r'<bpython-input-\d+>', filename):
+                    filename = '<input>'
+                    # strip linecache line number
                 value = SyntaxError(msg, (filename, lineno, offset, line))
                 sys.last_value = value
         list = traceback.format_exception_only(type, value)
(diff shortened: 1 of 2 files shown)


bpython/repl.py
```diff
@@ -149,9 +153,14 @@ def showtraceback(self):
...
-                for i, (fname, lineno, module, something) in enumerate(tblist):
+
+            for i, (fname, lineno, module, something) in enumerate(tblist):
+                # strip linecache line number
+                if re.match(r'<bpython-input-\d+>', fname):
+                    fname = '<input>'
+                    tblist[i] = (fname, lineno, module, something)
+                # Set the right lineno (encoding header adds an extra line)
+                if not py3:
                     if fname == '<input>':
```
thomasballinger: This was never true before, since it would be `<bpython-input-n>` instead.

bpython/test/test_interpreter.py
```diff
@@ -62,7 +55,7 @@ def f():
         def g():
             return f()
 
-        i.runsource('g()', encode=False)
+        i.runsource('g()')
```
thomasballinger: I've removed encode=False after coming to the conclusion that it should have been True the whole time: line numbers were being reported incorrectly in Python 2 previously, but now they're right. This test was passing before, but shouldn't have been.

PR #596, opened 2015-12-24
Simple eval attr access

thomasballinger:
Safely do attribute access to provide completion on simple expressions:
```
>>> import ast
>>> m = ast.parse('1 + 1')
>>> m.body[0].value.
┌───────────────────────────────────────────────────
│ col_offset left       lineno     op         right
└───────────────────────────────────────────────────
```


thomasballinger:
this completion in action: https://asciinema.org/a/32460

Issue #598, opened 2015-12-26
Can't display docstrings using unicode


thomasballinger:
>>> def foo():
    ...     "åß∂ƒ"
    ...
    >>> help(foo)
    Traceback (most recent call last):
      File "<input>", line 1, in <module>
        help(foo)
      File "/Users/tomb/Dropbox/code/bpython/bpython/curtsiesfrontend/_internal.py", line 56, in __call__
        return super(_Helper, self).__call__(*args, **kwargs)
      File "/Users/tomb/Dropbox/code/bpython/bpython/_internal.py", line 24, in __call__
        self.helper(*args, **kwargs)
      File "/usr/local/Cellar/python/2.7.10_2/Frameworks/Python.framework/Versions/2.7/lib/python2.7/pydoc.py", line 1794, in __call__
        self.help(request)
      File "/usr/local/Cellar/python/2.7.10_2/Frameworks/Python.framework/Versions/2.7/lib/python2.7/pydoc.py", line 1841, in help
        else: doc(request, 'Help on %s:')
      File "/usr/local/Cellar/python/2.7.10_2/Frameworks/Python.framework/Versions/2.7/lib/python2.7/pydoc.py", line 1578, in doc
        pager(render_doc(thing, title, forceload))
      File "/Users/tomb/Dropbox/code/bpython/bpython/curtsiesfrontend/_internal.py", line 49, in pager
        self._repl.pager(output)
      File "/Users/tomb/Dropbox/code/bpython/bpython/curtsiesfrontend/repl.py", line 1677, in pager
        tmp.write(text.encode(getpreferredencoding()))
    UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 39: ordinal not in range(128)
    >>>

PR #599, opened 2015-12-27
require unicode for pager

thomasballinger:
This should probably be looking at the the source for an encoding message, but I'm worried that by the time pydoc outputs this it's too late to find it.


thomasballinger:
Fixes https://github.com/bpython/bpython/issues/598

sebastinas:
Looks good.

Issue #597, opened 2015-12-25
 unsupported operand type(s) - windows python 3.5


alch3m1st:
> bpython-curses

```
  File "C:\Users\nikos\AppData\Local\Programs\Python\Python35\Scripts\bpython-curses-script.py", line 9, in <module>
    load_entry_point('bpython==0.14.2', 'console_scripts', 'bpython-curses')()
  File "c:\users\nikos\appdata\local\programs\python\python35\lib\site-packages\bpython\cli.py", line 1961, in main
    banner=banner)
  File "c:\users\nikos\appdata\local\programs\python\python35\lib\site-packages\bpython\cli.py", line 1843, in curses_wrapper
    return func(stdscr, *args, **kwargs)
  File "c:\users\nikos\appdata\local\programs\python\python35\lib\site-packages\bpython\cli.py", line 1880, in main_curses
    cols = make_colors(config)
  File "c:\users\nikos\appdata\local\programs\python\python35\lib\site-packages\bpython\cli.py", line 293, in make_colors
    ('W', 15),
TypeError: unsupported operand type(s) for +: 'dict_items' and 'list'
```

thomasballinger:
This is fixed by https://github.com/bpython/bpython/pull/569 which isn't in master - @alch3m1st if you'd like a workaround you could install master (`pip install git+git://github.com/bpython/bpython.git`)

We're hoping to do a release soon so this becomes the default.

PR #601, opened 2016-01-25
Fixing Windows Python2 compatibility

adityagupta679:
It breaks the code when using on windows python2

unsupported operand type | 'list' and 'list': python 2.7

### DIFF
--- a/bpython/cli.py
+++ b/bpython/cli.py
@@ -281,7 +281,7 @@ def make_colors(config):
     }
 
     if platform.system() == 'Windows':
-        c = dict(c.items() |
+        c = dict(list(c.items()) +
             [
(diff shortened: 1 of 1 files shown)


thomasballinger:
Introduced by https://github.com/bpython/bpython/issues/597

I don't know how big the Windows userbase is but this is probably enough to prompt a 15.1 soon.

adityagupta679:
Yes , also while we are on it , I think we need to modify the "contributing" documentation for windows developers .

thomasballinger:
Cool, a separate PR for that would be great.

adityagupta679:
I will get to it . I guess i will have to modify the contributing.rst file .

Issue #293, opened 2014-07-04
AttributeError: 'FakeStdin' object has no attribute 'encoding'


ikanobori:
`bpython`:

``` python
>>> import sys
>>> sys.stdin.encoding
Traceback (most recent call last):
  File "<input>", line 1, in <module>
AttributeError: 'FakeStdin' object has no attribute 'encoding'
```

`python` proper:

``` python
>>> import sys
>>> sys.stdin.encoding
'UTF-8'
```

---
- Bitbucket: https://bitbucket.org/bobf/bpython/issue/293
- Originally Reported By: [Steven Myint](http://bitbucket.org/myint)
- Originally Created At: 2014-06-27T04:38:10.625

PR #603, opened 2016-02-07
Emulate what "sys.stdout" does for "encoding"

myint:
This is similar to #293, which was fixed in a similar way. That case was for "sys.stdin".

Programs use this to check whether they need to encode a string before printing to standard out.


myint:
Never mind. This doesn't even get triggered under `bpython foo.py`, which was the case I was anticipating. I'm not sure if this is needed in interactive mode, so I'll close this.

PR #614, opened 2016-05-31
Fix TypeError when using F2 without some source

blueyed:
File "/usr/lib/python3.5/site-packages/bpython/cli.py", line 977, in p_key
        self.statusbar.message(_(e))
      File "/usr/lib/python3.5/site-packages/bpython/cli.py", line 1605, in message
        self.settext(s)
      File "/usr/lib/python3.5/site-packages/bpython/cli.py", line 1655, in settext
        if len(s) >= self.w:
    TypeError: object of type 'SourceNotFound' has no len()

I have used the same code as with `bpython/curtsiesfrontend/repl.py`.

### DIFF
--- a/bpython/cli.py
+++ b/bpython/cli.py
@@ -974,7 +974,7 @@ class CLIRepl(repl.Repl):
             try:
                 source = self.get_source_of_current_name()
             except repl.SourceNotFound as e:
-                self.statusbar.message(_(e))
+                self.statusbar.message(str(e))
             else:
                 if config.highlight_show_source:
                     source = format(PythonLexer().get_tokens(source),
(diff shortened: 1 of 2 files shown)


bpython/cli.py
```diff
@@ -974,7 +974,7 @@ def p_key(self, key):
             try:
                 source = self.get_source_of_current_name()
             except repl.SourceNotFound as e:
-                self.statusbar.message(_(e))
+                self.status_bar.message('%s' % (e, ))
```
sebastinas: Why not simply `str(e)`?

blueyed: That's what I had before, but then found the code in `bpython/curtsiesfrontend/repl.py` doing this.
Let's change both to use `str` then?

sebastinas:
Changing both to `str` sounds like a good plan.

blueyed:
Do you want that to be in two commits?

sebastinas:
I don't have a strong preference either way.

blueyed:
Done.

Issue #612, opened 2016-05-07
bpython interpreter with pwn module 


greenhand2learn:
I don't know is this problem has been asked. Here is my problem.
```
>>> from pwn import *
Traceback (most recent call last):
  File "<input>", line 1, in <module>
    from pwn import *
  File "/usr/lib/python2.7/dist-packages/bpython/curtsiesfrontend/repl.py", line 257, in load_module
    module = pkgutil.ImpLoader.load_module(self, name)
  File "/usr/lib/python2.7/pkgutil.py", line 246, in load_module
    mod = imp.load_module(fullname, self.file, self.filename, self.etc)
  File "/usr/local/lib/python2.7/dist-packages/pwn/__init__.py", line 2, in <module>
    from .toplevel import *
  File "/usr/local/lib/python2.7/dist-packages/pwn/toplevel.py", line 2, in <module>
    from pwnlib                  import *
  File "/usr/lib/python2.7/dist-packages/bpython/curtsiesfrontend/repl.py", line 257, in load_module
    module = pkgutil.ImpLoader.load_module(self, name)
  File "/usr/lib/python2.7/pkgutil.py", line 246, in load_module
    mod = imp.load_module(fullname, self.file, self.filename, self.etc)
  File "/usr/local/lib/python2.7/dist-packages/pwnlib/__init__.py", line 10, in <module>
    from . import \
  File "/usr/local/lib/python2.7/dist-packages/pwnlib/asm.py", line 45, in <module>
    from . import log
  File "/usr/local/lib/python2.7/dist-packages/pwnlib/log.py", line 69, in <module>
    from .term      import spinners, text
  File "/usr/local/lib/python2.7/dist-packages/pwnlib/term/__init__.py", line 1, in <module>
    from . import key, readline, text, termcap, keymap, term
  File "/usr/local/lib/python2.7/dist-packages/pwnlib/term/readline.py", line 2, in <module>
    from . import term, text
  File "/usr/local/lib/python2.7/dist-packages/pwnlib/term/text.py", line 111, in <module>
    sys.modules[__name__] = Module()
  File "/usr/local/lib/python2.7/dist-packages/pwnlib/term/text.py", line 22, in __init__
    self.num_colors = termcap.get('colors', default = 8)
  File "/usr/local/lib/python2.7/dist-packages/pwnlib/term/termcap.py", line 15, in get
    init()
  File "/usr/local/lib/python2.7/dist-packages/pwnlib/term/termcap.py", line 39, in init
[...]

PR #616, opened 2016-06-04
Added fileno support to FakeOutput

Caleb1994:
Some applications that use curses require that the object in `sys.stdout` be a file with the a method called `fileno`. In order to facilitate this, a `fileno` method was added to the FakeOutput object. An extra named parameter was added to the constructor as well to facilitate choosing the result of the `fileno` method. The default value is `1` (or standard output). 

The two allocations of FakeOutput (for stderr and stdout) were also changed to specify the file number to be returned by `fileno`.

These additions do not effect any existing code, but allow users to import libraries such as pwntools while using bpython.

### DIFF
--- a/bpython/curtsiesfrontend/coderunner.py
+++ b/bpython/curtsiesfrontend/coderunner.py
@@ -195,19 +195,29 @@ class CodeRunner(object):
 
 
 class FakeOutput(object):
-    def __init__(self, coderunner, on_write):
+    def __init__(self, coderunner, on_write, fileno=1):
         """Fakes sys.stdout or sys.stderr
 
         on_write should always take unicode
+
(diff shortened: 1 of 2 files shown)


Caleb1994:
This also fixes #612.

thomasballinger:
Thanks @Caleb1994. I guess now if a library writes to this file descriptor that we'll have display problems, but that seems better than the libs not being importable.

Note to myself: I wonder if putting a pipe there and letting things write to it and catching the data being written so we could integrate it would be useful? Would this break a lot of libs that expect a tty on stdout instead of a pipe?

sebastinas:
hrm, this has been rejected in the past for various reasons. Essentially since we have never figured out what it would mean to operate on the fileno.

PR #617, opened 2016-06-06
don't encode the file in "bpython file.py"

thomasballinger:
I'd like to do this to solve https://github.com/bpython/bpython/issues/608 but I'm never positive about encoding issues.

In Python 2 if we're reading a file from the command line, it's totally reasonable for us not to add our own encoding line, right? If it needs to be encoded, then there needed to be an encoding line there and if there is then we're munging it.


sebastinas:
I think we should first add tests of what we expect. Also, this change breaks the current tests.

thomasballinger:
Fix in https://github.com/bpython/bpython/pull/627

PR #618, opened 2016-06-07
completion box size changes based on terminal size

### LINKED ISSUE #466: max completions box height

thomasballinger:
Fix #466 

Try to preserve n lines of history (`try_preserve_history_height`) unless that space is needed to make the completion box m lines (`min_infobox_height`) long (provided there is content to fill it). If there's space in the terminal below the cursor to fill and infobox content to fill ti with, go ahead and fill it.

TODO:
  [x] - test desired behavior in each case
  [x] - docstring describing this layout technique

### DIFF
--- a/bpython/curtsiesfrontend/repl.py
+++ b/bpython/curtsiesfrontend/repl.py
@@ -1251,7 +1251,9 @@ class BaseRepl(BpythonRepl):
         self.current_stdouterr_line = ''
         self.stdin.current_line = '\n'
 
-    def paint(self, about_to_exit=False, user_quit=False):
+    def paint(self, about_to_exit=False, user_quit=False,
+              try_preserve_history_height=40,
+              min_infobox_height=4):
         """Returns an array of min_height or more rows and width columns, plus
         cursor position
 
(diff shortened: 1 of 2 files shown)


thomasballinger:
As @ikanobori suggested, I first tested the old behavior then checked it could be recreated in the old system, then changed the defaults to change behavior (see https://github.com/bpython/bpython/pull/618/files#diff-0f9439a6ef3e711e08826856e4ad77b8R699)

thomasballinger:
pinging @sebastinas, any thoughts on this? It
* more tests to previous behavior
* makes behavior more configurable
* changes behavior

Here's what the new behavior looks like: (gif is too fast, see https://asciinema.org/a/84660)

asciicast

thomasballinger:
This gif reveals some possible bugs, but they're in both versions. I've created a new issue for this: https://github.com/bpython/bpython/issues/634

Issue #625, opened 2016-07-27
Executing scripts with non ascii characters results in UnicodeDecodeError 


fzlee:
Hi team: bpython could be the best interpreter I have ever tried, which is really cool.

But I got some problems with it recently. I can reproduce this issue with the following conditions:

OS: OSX EI Capitan/Ubuntu 3.13.0-81-generic 64bit
bpython version 0.15 on top of Python 2.7.11 or 2.7.11

if you have a script like:

```
#!/usr/bin/env python2
# coding: utf-8
"""
    你好 # I put some non-ascii characters here  
    test.py
    ~~~~~~~~~~

"""

if __name__ == "__main__":
    pass
```

run `bpython -i test.py` and you will be able to see the following exceptions:

```
Traceback (most recent call last):
  File "/Users/xxx/.virtualenvs/virtualenv1/bin/bpython", line 11, in <module>
    sys.exit(main())
  File "/Users/xxx/.virtualenvs/virtualenv1/lib/python2.7/site-packages/bpython/curtsies.py", line 170, in main
    bpargs.exec_code(interp, exec_args)
  File "/Users/xxx/.virtualenvs/virtualenv1/lib/python2.7/site-packages/bpython/args.py", line 122, in exec_code
    interpreter.runsource(source, args[0], 'exec')
  File "/Users/xxx/.virtualenvs/virtualenv1/lib/python2.7/site-packages/bpython/repl.py", line 111, in runsource
    source = u'# coding: %s\n\n%s' % (self.encoding, source)
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe7 in position 330: ordinal not in range(128)
```

Removing non ascii characters from script and we can get the issue resolved, but this doesn't seem to be a good idea, since we sometimes really need those non ascii characters. Besides running this script with python works.

Seems bpython will append coding declaration automatically prior to code execution, and that could be the root cause:

```
# bpython/repl.py, line 108
source = u'# coding: %s\n\n%s' % (self.encoding, source)
```

Just wondering do we have any solutions regarding to this issue? Thank you in advance!

thomasballinger:
I've tried to write a patch for this, I'll write more tests for it soon.

fzlee:
Thank you, thomasballinger. this bug fix should help me a lot!

thomasballinger:
pending merging "Fix 625" #627

thomasballinger:
#627 is merged, assuming this is fixed

PR #627, opened 2016-07-29
Fix 625

thomasballinger:
fixing #625 by not adding encoding lines to source files.

### DIFF
--- a/bpython/inspection.py
+++ b/bpython/inspection.py
@@ -284,6 +284,7 @@ def is_callable(obj):
 
 
 get_encoding_re = LazyReCompile(r'coding[:=]\s*([-\w.]+)')
+get_encoding_line_re = LazyReCompile(r'^.*coding[:=]\s*[-\w.]+.*$')
 
 
 def get_encoding(obj):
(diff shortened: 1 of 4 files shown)


thomasballinger:
Add a new default value for `runsource(encode='auto')`, which uses whether the filename is None to determine whether input is coming from an interactive interface and therefore needs to have an encoding comment added.

Add test for every combination of (unicode/bytestring x encoding comment/none x encode=True/False/'auto'). Some combinations are invalid and raise ValueErrors.

bpython/inspection.py
```diff
@@ -284,6 +284,7 @@ def is_callable(obj):
 
 
 get_encoding_re = LazyReCompile(r'coding[:=]\s*([-\w.]+)')
+get_encoding_line_re = LazyReCompile(r'^.*coding[:=]\s*[-\w.]+.*$')
```
sebastinas: So why do we need that twice?

thomasballinger: I can combine these

sebastinas:
LGTM

Issue #622, opened 2016-07-06
AttributeError: 'FakeOutput' object has no attribute 'encoding'


goetzc:
`sys.stdout.encoding` gives this error.  
Also  `sys.stdin.encoding` gives `'UTF8'`, while python gives `'UTF-8'` for both cases.

[This](../pull/603) commit makes bpython work (naturally stdin is still different than in standard python):

``` python
bpython version 0.15 on top of Python 3.5.2 /usr/bin/python
>>> import sys
>>> sys.stdin.encoding
'UTF8'
>>> sys.stdout.encoding
'UTF-8'
```

thomasballinger:
Hm, I guess this should just be a property that returns whatever sys.stdout.encoding is, since that's how strings are actually being decoded. What are you using this for? (asking to try to understand if this is the right solution)

goetzc:
> I guess this should just be a property that returns whatever sys.stdout.encoding is

This is a much better solution!

>What are you using this for? 

I don't remember that :D

goetzc:
I found another AttributeError for 'FakeOutput', this time for 'fileno'. I was using [PyQtGraph.](https://github.com/pyqtgraph/pyqtgraph) Issue #628.

thomasballinger:
https://github.com/bpython/bpython/pull/629

gerryjenkinslb:
I ran into the problem in bpython import as well.
I would look at what idle does in its code base: 
see  cpython run.py PseudoFile class for example. if FakeOutput only encodes 'utf-8' then just return that.

rpdelaney:
Here's my +1:

```
$ python3
Python 3.7.2 (default, Feb 12 2019, 08:16:38) 
[Clang 10.0.0 (clang-1000.11.45.5)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import pipenv
>>> 
$ bpython
bpython version 0.17.1 on top of Python 3.7.2 /usr/local/opt/python/bin/python3.7
>>> import pipenv
Traceback (most recent call last):
  File "<input>", line 1, in <module>
    import pipenv
  File "/Users/ryan.delaney/Library/Python/3.7/lib/python/site-packages/pipenv/__init__.py", line 47, in <module>
    from .cli import cli
  File "/Users/ryan.delaney/Library/Python/3.7/lib/python/site-packages/pipenv/cli/__init__.py", line 3, in <module>
    from .command import cli
[...]

PR #629, opened 2016-08-05
Fix fd encodings

thomasballinger:
I propose we provide `fileno()` methods and `.encoding` properties on fake file objects that replace sys.stdin/stdout/stderr. Applications may use them in ways that work or in ways that don't, but we should just provide them and see what happens.

Pro: users can use more libraries with bpython

Con: code may break in less expected ways that the straightforward attribute errors they break on now, e.g. the display may break when a program writes directly to sys.stdout.fileno() because that won't be caught by bpython

### DIFF
--- a/bpython/test/test_curtsies_coderunner.py
+++ b/bpython/test/test_curtsies_coderunner.py
@@ -52,4 +52,4 @@ class TestFakeOutput(unittest.TestCase):
 
     def test_bytes(self):
         out = FakeOutput(mock.Mock(), self.assert_unicode, None)
-        out.write('native string type')
+        out.write("native string type")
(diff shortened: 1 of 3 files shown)


thomasballinger:
Given more activity on https://github.com/bpython/bpython/issues/622 I'll plan to look at merging this tomorrow.

bpython/test/test_curtsies_coderunner.py
```diff
@@ -51,5 +51,5 @@ def assert_unicode(self, s):
         self.assertIsInstance(s, type(u""))
 
     def test_bytes(self):
-        out = FakeOutput(mock.Mock(), self.assert_unicode)
-        out.write("native string type")
+        out = FakeOutput(mock.Mock(), self.assert_unicode, None)
+        out.write('native string type')
```
sebastinas, line 55: Please change it back to `""` (or run black before merging the changes).

PR #636, opened 2016-09-29
Adding CrazyLocals for importing stuff automatically

eliasdorneles:
So, this is a really bad idea.

This makes names from sys.modules magically available without
needing to import them.

It does that by replacing the locals dict by a dict subclass that looks
up names in sys.modules when they're not present in current locals.

Please do NOT merge!

But have fun. :)

### DIFF
--- a/bpython/curtsiesfrontend/interpreter.py
+++ b/bpython/curtsiesfrontend/interpreter.py
@@ -61,6 +61,21 @@ class BPythonFormatter(Formatter):
         outfile.write(parse(o.rstrip()))
 
 
+class CrazyLocals(dict):
+    def __getitem__(self, name):
+        try:
+            return super(CrazyLocals, self).__getitem__(name)
+        except KeyError:
+            import sys
+            try:
+                return sys.modules[name]
+            except KeyError:
+                for mod in sys.modules.values():
+                    if hasattr(mod, name):
+                        return getattr(mod, name)
+                raise
+
+
 class Interp(ReplInterpreter):
     def __init__(self, locals=None, encoding=None):
         """Constructor.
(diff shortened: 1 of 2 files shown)


thomasballinger:
I'm surprised and confused about this working in Python 2!

I'm sort of curious about the failing tests in Python 2.7, I glanced at them and I didn't understand what was broken. No pressure for this not-going-to-be-merged PR, but if you look at them maybe leave a comment about what broke.

eliasdorneles:
I was just looking into the tests, the problem is that some expectations fail because some tests are expecting some names to not be defined.

For `test_traceback` I could see what's going on: it expects to see a NameError when calling `g()` in a new interpreter, but the actual exception is `TypeError: 'int' object is not callable` -- which I figure means some module available in the path during test run has a global `g` constant. :)

For the other errors, they're trying to get the line with error by causing a NameError using `abc`. But with crazy locals, `abc` gives you `<module 'abc' from '/usr/lib/python2.7/abc.pyc'>`. :metal: 

I'm pushing a fix changing the names `abcd` and `gfunc`. :)

thomasballinger:
Ah cool, that all makes sense. Wow, thanks for changing all those tests. I wonder if we should merge the changes to tests anyway since this is a conceivable way that things could fail?

It probably goes without saying that this isn't getting merged. What about as a module? The changes you make in `bpython/curtsiesfrontend/interpreter:Interpreter` could be monkeypatched in, so this could be a `bpython-autoimport` plugin of sorts.

eliasdorneles:
Hm, I don't know, doesn't it seem like a maintenance headache? :)

I'll send a second PR with the test changes.

thomasballinger:
@eliasdorneles sure, but only a maintenance headache for you, not bpython!

eliasdorneles:
Haha, right!
I'll see how to make a module, just for fun! :)

thomasballinger:
Closing to reduce confusion

Issue #631, opened 2016-08-18
add operate-and-get-next (C-o)


thomasballinger:
As it works in [bash](http://linux.die.net/man/1/bash). If we're not getting multiline history anytime soon, this will help a bit.

PeriGK:
Hi @thomasballinger,

I am afraid I dont quite understand the requirement? could you please provide me with a usage scenario?

Thanks,
P.

thomasballinger:
Hi @PeriGK, sure! Thanks for your interest in the issue, please let me know if you want any help with this.

When you enter a multiline command, or several lines that go together, and then want to run them again, right now you have to hit up arrow enough times to get back to that line in history, then go back that many times again to get the next line, then go back again, once for each line. Ctrl-o

Ctrl-o gets swallowed in bash on OSX by default, so if you're on a mac and want to try it out try running `stty -iexten` in the terminal first or ssh into a linux server to see it.

Here's a [CPython issue about adding it](https://bugs.python.org/issue22228), looks like it hasn't been merged yet. This isn't really related to us trying to implement it, it just might be a useful description of the feature.
 
[this asciinema shows use of ctrl-o](https://asciinema.org/a/90139)

I'm imagining we add a key to `bpython/curtsiesfrontend/repl.py` that is the same as hitting entry for the current command, except that the current history index isn't reset back to the end.

PeriGK:
Hi @thomasballinger ,

Thanks for the hints. Ok I think I can start working on the bug. 

Could you please let me know how do we understand a multiline command? Will it be that it just ends with "\"?

Here is my understanding till now, please let me know if I am totally off with something:
- I will need to go through the history list(self.history), lets say backwards.
- When the first "\" or any other multiline indicator is seen, I will be creating a string till "\" is not there. - Of course, I will concatenate the last-before-the-first-backslash-string seen, as this is the last line of the multiline command.
- All this work(or at least most of it) will be happening inside BaseRepl class. The only way to execute the above logic, will be to hit Ctrl+o in the keyboard.

Thanks,
P.

thomasballinger:
Yep, I agree BaseRepl is the place for this. However, this does not require detecting multiline commands. Ctrl-o doesn't treat lines that are part of mutliline commands any differently than single line commands, it only runs the current line and brings up the next command in history.

You might try out ctrl-o to get a feel for how it works. Here's an online shell (type `bash` first) that you can use to try this out: https://www.tutorialspoint.com/unix_terminal_online.php

[...]

PR #660, opened 2016-12-10
Get next command with ctrl+o(issue 631)

PeriGK:
Hi @thomasballinger ,

As discussed, this is an incomplete PR about issue #631 

I can retrieve the next-to-be-shown-line(stored in variable next). I was expect setting the current line to the new one, to work fine and print it. Am I missing something?

Also, apart from print commands, is there any better way to use a debugger here, given that I am running it with python -m?

Thanks,
P.

### DIFF
--- a/bpython/curtsiesfrontend/repl.py
+++ b/bpython/curtsiesfrontend/repl.py
@@ -676,7 +676,7 @@ class BaseRepl(BpythonRepl):
         elif e in ("<Ctrl-d>",):
             self.on_control_d()
         elif e in ("<Ctrl-o>",):
-            self.on_control_o()
+            self.operate_and_get_next()
         elif e in ("<Esc+.>",):
             self.get_last_word()
         elif e in key_dispatch[self.config.reverse_incremental_search_key]:
@@ -846,19 +846,24 @@ class BaseRepl(BpythonRepl):
             self.current_line = (self.current_line[:self.cursor_offset] +
                                  self.current_line[(self.cursor_offset + 1):])
 
-    def on_control_o(self):
-        next_idx = -self.rl_history.index + 1
-        next = self.rl_history.entries[next_idx]
-        self.on_enter()
-        print("Next is : " + next)
-        self._set_current_line(next)
-
     def cut_to_buffer(self):
         self.cut_buffer = self.current_line[self.cursor_offset:]
         self.current_line = self.current_line[:self.cursor_offset]
 
     def yank_from_buffer(self):
         pass
+
+    def operate_and_get_next(self):
+        # If we have not navigated back in history
+        # ctrl+o will have the same effect as enter
+        if self.rl_history.index == 0 or self.rl_history.index == 1:
+            self.on_enter()
+            return
+
+        index = self.rl_history.index
+        self.on_enter()
+        self.rl_history.index = index
+        self._set_current_line(self.rl_history.entries[-index], reset_rl_history=False)
 
     def up_one_line(self):
         self.rl_history.enter(self.current_line)
--- a/bpython/curtsiesfrontend/repl.py
+++ b/bpython/curtsiesfrontend/repl.py
@@ -970,7 +970,6 @@ class BaseRepl(BpythonRepl):
         if len(char) > 1 or is_nop(char):
             return
         if self.incr_search_mode:
-            print("Incr search mode")
             self.add_to_incremental_search(char)
         else:
             self._set_current_line((self.current_line[:self.cursor_offset] +
--- a/bpython/curtsiesfrontend/repl.py
+++ b/bpython/curtsiesfrontend/repl.py
@@ -788,9 +788,11 @@ class BaseRepl(BpythonRepl):
         else:
             self.cut_buffer = cut
 
-    def on_enter(self, insert_into_history=True):
+    def on_enter(self, insert_into_history=True, reset_rl_history=True):
         # so the cursor isn't touching a paren TODO: necessary?
         self._set_cursor_offset(-1, update_completion=False)
+        if reset_rl_history:
+            self.rl_history.reset()
 
         self.history.append(self.current_line)
         self.push(self.current_line, insert_into_history=insert_into_history)
@@ -856,14 +858,7 @@ class BaseRepl(BpythonRepl):
     def operate_and_get_next(self):
         # If we have not navigated back in history
         # ctrl+o will have the same effect as enter
-        if self.rl_history.index == 0 or self.rl_history.index == 1:
-            self.on_enter()
-            return
-
-        index = self.rl_history.index
-        self.on_enter()
-        self.rl_history.index = index
-        self._set_current_line(self.rl_history.entries[-index], reset_rl_history=False)
+        self.on_enter(reset_rl_history=False)
 
     def up_one_line(self):
         self.rl_history.enter(self.current_line)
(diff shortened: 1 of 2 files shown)


bpython/curtsiesfrontend/repl.py
```diff
@@ -844,6 +846,13 @@ def on_control_d(self):
             self.current_line = (self.current_line[:self.cursor_offset] +
                                  self.current_line[(self.cursor_offset + 1):])
 
+    def on_control_o(self):
+        next_idx = -self.rl_history.index + 1
+        next = self.rl_history.entries[next_idx]
+        self.on_enter()
```
thomasballinger: FYI `on_enter()` has the side effect of resetting the current `rl_history.index` to 0

bpython/curtsiesfrontend/repl.py
```diff
@@ -844,6 +846,13 @@ def on_control_d(self):
             self.current_line = (self.current_line[:self.cursor_offset] +
                                  self.current_line[(self.cursor_offset + 1):])
 
+    def on_control_o(self):
```
thomasballinger: We probably want an escape hatch at the beginning of this method like
~~~
if self.rl_history.index == 0:
    self.on_enter()
    return
~~~
so ctrl-o works like enter if we aren't back in history.

thomasballinger: Actually maybe the same escape hatch should apply to `self.rl_history.index == 1` as well, because this math doesn't work in that case either (the case where we're looking at the last entry in history).

bpython/curtsiesfrontend/repl.py
```diff
@@ -844,6 +846,13 @@ def on_control_d(self):
...
+        next_idx = -self.rl_history.index + 1
+        next = self.rl_history.entries[next_idx]
+        self.on_enter()
+        print("Next is : " + next)
+        self._set_current_line(next)
```
thomasballinger: The current code works well the first time, but not the second time; if trying to use ctrl-o to quickly re-enter a function longer than 2 lines for example. We should reset `self.rl_history.index` here too so that we're still back in history. (only a problem because `on_enter()` sets `self.rl_history.index` to 0).

Maybe something like
~~~
index = self.rl_history.index
...everything else in the method
self.rl_history.index = index
~~~

bpython/curtsiesfrontend/repl.py
thomasballinger: Hm maybe we should fix the math instead, because in bash ctrl-o does work with the last line of history. You could change this by never doing `next_idx = -self.rl_history.index + 1`, and instead using the current index as described in the bottom comment - something like
~~~
def on_control_o(self):
    index = self.rl_history.index
    self.on_enter()  # this resets self.rlu_history.index
    self.rlu_history.index = index
    self._set_current_line(self.rlu_history.entries[-index))
~~~

bpython/curtsiesfrontend/repl.py
thomasballinger: OH one more relevant note: `self._set_current_line` will reset `rl_history.index` as well, so be sure to set the index after this, or use `self._set_current_line(next, reset_rl_history=False)`

thomasballinger:
Hmm good question re debugging. I've used [winpdb](http://winpdb.org/) for remote debugging with bpython before, but usually I'll just use logging:
* add code like `logger.info('hello!')`
* run bpython with `python -m bpython -L`
* in another terminal, run `tail -f bpython.log`

Another possibly useful debugging trick: you can get the current repl object in bpython with
~~~
from bpython.curtsies import repl
~~~
and try calling methods there.

PeriGK:
Thanks for the feedback Thomas. I have followed your suggestions and I am experiencing something kinda weird. If I hit ctrl+o on an assignment command it works fine. 

But if I do it on a print command it's not. It seems to have a similar effect to "resetting rl_history.index" but I am not sure, why this would happen(it would probably cause more severe issues in such case). Anyway, I have pushed again and will keep investigating.

Thanks,
P.

thomasballinger:
Oh good catch, this doesn't work if any output is generated. It looks like `        self.run_code_and_maybe_finish()` is run multiple times for many commands, including those that generate output, and every time it is run after the first time is after we set `rl_history.index` to the saved value.

Hm, I think we need another approach then. I made the following changes to your code and things seem ok:

~~~
diff --git a/bpython/curtsiesfrontend/repl.py b/bpython/curtsiesfrontend/repl.py
index f46bfcb..6e6be46 100644
--- a/bpython/curtsiesfrontend/repl.py
+++ b/bpython/curtsiesfrontend/repl.py
@@ -788,9 +788,11 @@ class BaseRepl(BpythonRepl):
         else:
             self.cut_buffer = cut
 
-    def on_enter(self, insert_into_history=True):
+    def on_enter(self, insert_into_history=True, reset_rl_history=True):
         # so the cursor isn't touching a paren TODO: necessary?
         self._set_cursor_offset(-1, update_completion=False)
+        if reset_rl_history:
+            self.rl_history.reset()
 
         self.history.append(self.current_line)
         self.push(self.current_line, insert_into_history=insert_into_history)
@@ -856,14 +858,7 @@ class BaseRepl(BpythonRepl):
     def operate_and_get_next(self):
         # If we have not navigated back in history
         # ctrl+o will have the same effect as enter
-        if self.rl_history.index == 0 or self.rl_history.index == 1:
-            self.on_enter()
-            return
-
-        index = self.rl_history.index
-        self.on_enter()
-        self.rl_history.index = index
-        self._set_current_line(self.rl_history.entries[-index], reset_rl_history=False)
+        self.on_enter(reset_rl_history=False)
 
     def up_one_line(self):
         self.rl_history.enter(self.current_line)
@@ -1091,7 +1086,11 @@ class BaseRepl(BpythonRepl):
                     self.current_stdouterr_line, self.width))
                 self.current_stdouterr_line = ''
 
-            self._set_current_line(' ' * indent, update_completion=True)
+            if self.rl_history.index == 0:
+                self._set_current_line(' ' * indent, update_completion=True)
+            else:
+                self._set_current_line(self.rl_history.entries[-self.rl_history.index],
+                                       reset_rl_history=False)
             self.cursor_offset = len(self.current_line)
 
     def keyboard_interrupt(self):
~~~
I think this would be a promising direction: reset the history index in `on_enter`, except in our ctrl-o case where we don't want it reset, and then when a command finishes bring up the current entry in rl_history.

If this seems good, the other thing to fix would be to try to use some of the convenient methods in bpython.history.History instead of accessing `.index` manually.

PeriGK:
Hi @thomasballinger 

I have followed your suggestions and have pushed some more changes. I have tested them and they look fine. Let me know if you need anything else.

thanks,
P.

thomasballinger:
Sorry for the delay, busy moving today but I can look at this tomorrow.

bpython/curtsiesfrontend/repl.py
```diff
@@ -961,6 +970,7 @@ def add_normal_character(self, char):
         if len(char) > 1 or is_nop(char):
             return
         if self.incr_search_mode:
+            print("Incr search mode")
```
sebastinas: This seems to be left-over debugging output.

PeriGK: Oops, my bad :) I have removed it now.

PeriGK:
H guys,

Just checking, out of interest, as I realised this is still open...everything ok with this PR? Is there anything else needed? :)

thomasballinger:
Merged (PR #667), thanks @PeriGK

PeriGK:
Hello @thomasballinger 

Is this indeed merged? I tried to use it on the 0.16 version and it is not operating. Also, I couldnt find it in the changelog and thought to ask. :)

Happy Christmas,
P.

thomasballinger:
@PeriGK Does this work for you in bash? If e.g. you're on a mac, you'll need to add ``stty discard undef` to your bashrc to avoid it being swallowed by the terminal driver.

PeriGK:
@thomasballinger works fine for me. I tested using my branch(it is a bit old though)

thomasballinger:
@PeriGK sorry that wasn't clear, I mean it's the released version - can you download the newest bpython and try it out?

PR #661, opened 2016-12-12
Modified repl.py to conform to follow more PEP standards including 8 and 263

shanahanjrs:
This merge includes style and indentation fixes, removes unused function args, and updates variable names to be more descriptive and improve code readability.

### DIFF
--- a/bpython/repl.py
+++ b/bpython/repl.py
@@ -1,27 +1,29 @@
-# encoding: utf-8
+# -*- coding: utf-8 -*-
 
-# The MIT License
-#
-# Copyright (c) 2009-2011 the bpython authors.
-# Copyright (c) 2012-2013,2015 Sebastian Ramacher
-#
-# Permission is hereby granted, free of charge, to any person obtaining a copy
-# of this software and associated documentation files (the "Software"), to deal
-# in the Software without restriction, including without limitation the rights
-# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
-# copies of the Software, and to permit persons to whom the Software is
-# furnished to do so, subject to the following conditions:
-#
-# The above copyright notice and this permission notice shall be included in
-# all copies or substantial portions of the Software.
-#
-# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
-# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
-# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
-# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
-# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
-# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
-# THE SOFTWARE.
+"""
+The MIT License
+
+Copyright (c) 2009-2011 the bpython authors.
+Copyright (c) 2012-2013,2015 Sebastian Ramacher
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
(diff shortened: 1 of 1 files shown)


bpython/repl.py
```diff
@@ -1,27 +1,29 @@
...
+all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+"""
```
sebastinas: Why? We consistently use #-commented license headers, so please do not change that.

bpython/repl.py
```diff
@@ -1099,15 +1103,15 @@ def open_in_external_editor(self, filename):
...
+        if not os.path.isfile(self.config.config_path):
             if self.interact.confirm(_("Config file does not exist - create "
                                        "new from default? (y/N)")):
                 try:
                     default_config = pkgutil.get_data('bpython',
                                                       'sample-config')
                     if py3:  # py3 files need unicode
                         default_config = default_config.decode('ascii')
-                    bpython_dir, script_name = os.path.split(__file__)
+                    #bpython_dir, script_name = os.path.split(__file__)
```
sebastinas: Just remove it if `bpython_dir` and `script_name` are unused.

bpython/repl.py
```diff
@@ -56,6 +58,7 @@
 
 
 class RuntimeTimer(object):
+    """Calculate running time"""
```
thomasballinger: I like docstrings, maybe we can do one better:
`"""Records cumulative time spend in this context manager"""`

bpython/repl.py
```diff
@@ -601,7 +604,7 @@ def get_args(self):
                 try:
                     fake_cursor = self.current_line.index(func) + len(func)
                     f = simpleeval.evaluate_current_attribute(
-                            fake_cursor, self.current_line, self.interp.locals)
+                        fake_cursor, self.current_line, self.interp.locals)
```
thomasballinger: I'm not very familiar with these rules, is this recommended by PEP8? I interpret https://www.python.org/dev/peps/pep-0008/#indentation as saying the original is correct.

bpython/repl.py
```diff
@@ -770,13 +773,13 @@ def line_is_empty(line):
             indentation = 0
         return indentation
 
-    def formatforfile(self, s):
+    def formatforfile(self, stdout):
```
thomasballinger: I agree `s` isn't great, but at least there's the convention that `s` is a string. Stdout sounds like a file descriptor or sys.stdout or something to me. How about `session_output`?

bpython/repl.py
```diff
@@ -817,11 +820,11 @@ def write2file(self):
                 self.interact.notify(_('Save cancelled.'))
                 return
 
-        s = self.formatforfile(self.getstdout())
+        stdout_text = self.formatforfile(self.getstdout())
```
thomasballinger: This isn't just stdout text, it's session text; the interleaving of stdin, stdout, and stderr.

bpython/repl.py
thomasballinger: Or `Tracks cumulative time spend in this context`

bpython/repl.py
```diff
@@ -1069,8 +1072,9 @@ def tokenize(self, s, newline=False):
     def clear_current_line(self):
         """This is used as the exception callback for the Interpreter instance.
         It prevents autoindentation from occurring after a traceback."""
+        # XXX: Empty function
```
thomasballinger: Is this a PEP or a pylint thing? Why is this useful?

bpython/repl.py
shanahanjrs: Sure those sound much better! Ill make the changes you purposed and resubmit my pull request.

bpython/repl.py
Trundle: It would also change the license header from a comment to the docstring of the `bpython.repl` module.

sebastinas:
Squashed the two commits and rebased them on current master. Thanks.

PR #667, opened 2017-02-12
Changes to PeriGK's ctrl-o PR

thomasballinger:
I tried something a little different but it's been so long since I last looked at it I didn't want to merge it right away.


thomasballinger:
Thanks so much @PeriGK for your work on this, and apologies for the slow turnaround.

PR #679, opened 2017-04-18
Allow bpython/bpdb outside main thread

### LINKED ISSUE #555: Let bpython be run from not main thread
I think this is just a problem with signal handlers. I'm trying playing with this now.

maxnordlund:
Because [signal.signal][1] is only allowed to be called from the main thread, if you use bpdb in another thread it will crash upon entering the repl.  By ignoring the resulting `ValueError` the show can go on.

This makes bpdb useful in Django's runserver, even with threading and/or live reload on.

A question is what to tell the user when this happens? Just ignore it like here, or log something? What on which level?

Feedback much appreciated.

Closes #555 I think

[1]: https://docs.python.org/3/library/signal.html#signal.signal

### DIFF
--- a/bpython/curtsiesfrontend/repl.py
+++ b/bpython/curtsiesfrontend/repl.py
@@ -11,6 +11,7 @@ import signal
 import subprocess
 import sys
 import tempfile
+import threading
 import time
 import unicodedata
 from six.moves import range
@@ -529,12 +530,10 @@ class BaseRepl(BpythonRepl):
         self.orig_sigwinch_handler = signal.getsignal(signal.SIGWINCH)
         self.orig_sigtstp_handler = signal.getsignal(signal.SIGTSTP)
 
-        try:
+        if isinstance(threading.current_thread(), threading._MainThread):
+            # This turns off resize detection and ctrl-z suspension.
             signal.signal(signal.SIGWINCH, self.sigwinch_handler)
             signal.signal(signal.SIGTSTP, self.sigtstp_handler)
-        except ValueError:
-            pass # Ignore "signal only works in main thread"
-            # This turns off resize detection and ctrl-z suspension.
 
         self.orig_meta_path = sys.meta_path
         if self.watcher:
@@ -548,12 +547,10 @@ class BaseRepl(BpythonRepl):
         sys.stdout = self.orig_stdout
         sys.stderr = self.orig_stderr
 
-        try:
+        if isinstance(threading.current_thread(), threading._MainThread):
+            # This turns off resize detection and ctrl-z suspension.
             signal.signal(signal.SIGWINCH, self.orig_sigwinch_handler)
             signal.signal(signal.SIGTSTP, self.orig_sigtstp_handler)
-        except ValueError:
-            pass # Ignore "signal only works in main thread"
-            # This turns off resize detection and ctrl-z suspension.
 
         sys.meta_path = self.orig_meta_path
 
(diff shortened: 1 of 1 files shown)


thomasballinger:
Cool, thanks @maxnordlund. I think we should use this, but first we should document what stops working when these signal handlers can't be installed. Hm yeah maybe we should let the user know too? A startup message that says what won't work? I can plug that in; most important to me is getting a code comment that says what won't work, but agreed a user message would be nice.

maxnordlund:
Yeah, but I couldn't really figure out what it turns off. I guess `SIGWINCH` is for resizing, but what doe the `SIGTSTP ` do?

thomasballinger:
It's backgrounding the process, so with that off I assume ctrl-z will stop
working, not positive.

maxnordlund:
Yes it would seem that's what they are for, judging from the commit headlines in [git blame](https://github.com/bpython/bpython/blame/master/bpython/curtsiesfrontend/repl.py#L549-L565).

I'll add a comment about that, and this should be good to go?

thomasballinger:
Yep, I'm ok with this! I wonder about some things (will behavior be weird in this state) but you're just making something that didn't work before work. I'll look at a user message later.

thomasballinger:
Please play around with it a little (not a blocker), since we don't have tests for this signal hehavior, it's just manual testing. Hope it works well for you in Django.

maxnordlund:
Before I made the pull request I did this on a Django project at work, and it did work there. Though I monkey-patched the signal module directly because it was easier then trying to monkey-patch bpython- Of course that doesn't feel right, hence this PR.

```python
import signal
original = signal.signal

def patched_signal(signalnum, handler):
    try:
        original(signalnum, handler)
    except ValueError:
        pass

setattr(signal, "signal", patched_signal)
```

sebastinas:
From what I understand from `signal.signal`s documentation, `ValueError`s may also be raised in other cases. So please check if the `ValueError` really comes from this one specific case.

thomasballinger:
You might check that we're on the main thread and if not not try since that
would be clearer

maxnordlund:
> From what I understand from signal.signals documentation, ValueErrors may also be raised in other cases. So please check if the ValueError really comes from this one specific case.

The only other case concerns Windows, https://docs.python.org/3/library/signal.html#signal.signal and since bpython uses SIGWINCH it doesn't work on Windows (probably for other reasons as well).

> You might check that we're on the main thread and if not not try since that
would be clearer

I thought it was more pythonic to assume the happy path and catch exceptions if it's not. E.g. not check if a method exists, just try to call it. Follow the duck-typing-way or something like that.

maxnordlund:
I switched to an if statement, as suggested by @thomasballinger, but there be dragons there.

In python 3 we have the [threading.main_thread](https://docs.python.org/3.6/library/threading.html#threading.main_thread) function, but in python 2 you basically have to resort to [some form of hack](https://stackoverflow.com/questions/23206787/check-if-current-thread-is-main-thread-in-python).

@sebastinas I've been thinking about your question a bit more and even if the try-except catches more ValueErrors it's fine because the whole point of this PR is to make sure we don't crash.

maxnordlund:
Ping @thomasballinger, how do we move forward with this? I personally like the try/except version better, as it feels less hacky. But it's you call.

maxnordlund:
@thomasballinger ping

thomasballinger:
I just change Curtsies to be fine with this too, trying it out with this now.

thomasballinger:
Looks like this needs to happen in coderunner as well

thomasballinger:
Still to do:
* focus_on_subprocess
* ?

thomasballinger:
This seems not to be done (at least not in 0.17.1)

thomasballinger:
hm now it's working for me, not sure what's up

maxnordlund:
It can be a race condition between when the signal handler is being registered, if it happens to be on the main thread it's fine. It's just when you try to register a handler on some other thread Python complains (and POSIX as well I guess).

PR #680, opened 2017-04-22
don't ignore rc flines by by overwriting rcLines

### LINKED ISSUE #677: Support .pdbrc in bpdb (or variation)
I was trying to add autocomplete to bpdb, I know it already works in bpython.

See https://github.com/python-git/python/blob/master/Lib/pdb.py#L78-L95 for `.pdbrc` implementation

I'm wondering why these lines aren't in https://github.com/bpython/bpython/blob/master/bpdb/debugger.py

Was there a reason this isn't supported? Possibly with it's own .rc eg `.bpdbrc`

So I just stepped through the code and since bpdb extends pdb, it is loading `.pdbrc`, but it's effects aren't seen in bpdb, at least for me.

Ah, line 36 in bpdb/debugger.py sets `self.rcLines = []` which means that `.pdbrc` won't be loaded.

My .pdbrc contains:

```python
import rlcompleter

print('pdb autocomplete')
import pdb  # just in case
pdb.Pdb.complete = rlcompleter.Completer(locals()).complete
```

Commenting out line 36 from dpdb/debugger.py properly loads my `.pdbrc` Can this line be commented out, or a variation of pdb.py's code be added to load a `.bpdbrc`?

I'm on Windows 10 using MinGW64 provided by git, Python 3.5, bpython 0.14.2

thomasballinger:
fix #677

### DIFF
--- a/bpdb/debugger.py
+++ b/bpdb/debugger.py
@@ -33,7 +33,6 @@ class BPdb(pdb.Pdb):
 
     def __init__(self, *args, **kwargs):
         pdb.Pdb.__init__(self, *args, **kwargs)
-        self.rcLines = []
         self.prompt = '(BPdb) '
         self.intro = 'Use "B" to enter bpython, Ctrl-d to exit it.'
 


sebastinas:
Shouldn't we use our own config file for bpdb?

thomasballinger:
Hm maybe, it feels like it's so similar to pdb that people would just want to use that. Maybe we could overwrite the `.pdbrc` options with our own?

thomasballinger:
Happy to roll this back, I thought about that but couldn't think of a reason people would need a separate rc for bpdb since it was so similar.

sebastinas:
Sure, it's similar, but it's not the same. It's a different entry point after all. We could do something like the following: prefer `$XDG_CONFIG_HOME/bpython/bpdb` if it exists, and if not, fall back to `.pdbrc`. The manpage should also document the rc files.

PR #690, opened 2017-07-08
Add default_autoreload config option

afrieder:
This is something I personally would love to have on by default and I'm sure others feel the same way. No harm in adding it as an option.


thomasballinger:
@afrieder what would this do?

thomasballinger:
Oh ha sorry, this is a pull request, I see! That's interesting.

afrieder:
Yeah, I really really love the autoreload feature, but I find myself turning it on every single time I open bpython (since I mostly use it for rapid prototyping/debugging of small scripts), so it'd be helpful to just enable it by default.

PR #694, opened 2017-07-26
simple fix (ctrl-c handling probably doesn't work in all cases)

thomasballinger:
Follow on to https://github.com/bpython/bpython/pull/679

Allows code to be run not in the main thread

### DIFF
--- a/bpython/curtsiesfrontend/coderunner.py
+++ b/bpython/curtsiesfrontend/coderunner.py
@@ -13,6 +13,7 @@ makes more sense to you.
 """
 
 import code
+import threading
 import signal
 import greenlet
 import logging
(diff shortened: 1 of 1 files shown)


maxnordlund:
Looks good I think, it's basically the same as the previous PR. But I wonder if it's a good idea to put this inside an internal signal package to hide this dance. Then just update imports and all _should_ be good: `import internal.signal as signal`

bpython/curtsiesfrontend/coderunner.py
```diff
@@ -130,16 +131,19 @@ def run_code(self, for_code=None):
         if source code is complete, returns "done"
         if source code is incomplete, returns "unfinished"
         """
+        is_main_thread = isinstance(threading.current_thread(), threading._MainThread)
```
sebastinas: Please reuse `is_main_thread` from `bpython.curtsiesfrontend.repl`.

PR #702, opened 2017-08-19
fix encoding issues in urwid and curses

### LINKED ISSUE #701: bpython-urwid on python 2 can't format tracebacks that contain non-ASCII characters
Steps to reproduce:

1. `pip install bpython-urwid`
1. run `bpython-urwid`
1. paste `é`
1. hit enter

expected result:

A traceback about é, then another Python prompt.

Actual result:

```
  File "./Library/Python/2.7/bin/bpython-urwid", line 11, in <module>
    sys.exit(main())
  File ".../bpython/urwid.py", line 1317, in main
    myrepl.main_loop.screen.run_wrapper(run_with_screen_before_mainloop)
  File ".../urwid/display_common.py", line 764, in run_wrapper
    return fn()
  File ".../bpython/urwid.py", line 1268, in run_with_screen_before_mainloop
    myrepl.main_loop.run()
  File ".../urwid/main_loop.py", line 278, in run
    self._run()
  File ".../urwid/main_loop.py", line 376, in _run
    self.event_loop.run()
  File ".../urwid/main_loop.py", line 682, in run
    self._loop()
  File ".../urwid/main_loop.py", line 719, in _loop
    self._watch_files[fd]()
  File ".../urwid/raw_display.py", line 393, in <lambda>
    event_loop, callback, self.get_available_raw_input())
  File ".../urwid/raw_display.py", line 493, in parse_input
    callback(processed, processed_codes)
  File ".../urwid/main_loop.py", line 403, in _update
    self.process_input(keys)
  File ".../urwid/main_loop.py", line 509, in process_input
    something_handled |= bool(self.unhandled_input(k))
  File ".../urwid/main_loop.py", line 555, in unhandled_input
    return self._unhandled_input(input)
  File ".../bpython/urwid.py", line 1039, in handle_input
    self.stdout_hist += inp.encode(locale.getpreferredencoding())
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 0: ordinal not in range(128)
```

and the interpreter exits.

thomasballinger:
fixes #701 

Make it clear with comments that self.stdout_hist is bytes in Python 2, unicode in Python 3. Blech, would be nice to fix that to be unicode everywhere.

To debug I made self.stdout_hist a type-checking property, maybe this is worth doing for real? Also maybe static typechecking would be nice? But this PR just fixes this bug.

### DIFF
--- a/bpython/cli.py
+++ b/bpython/cli.py
@@ -330,7 +330,7 @@ class CLIRepl(repl.Repl):
         repl.Repl.__init__(self, interp, config)
         self.interp.writetb = self.writetb
         self.scr = scr
-        self.stdout_hist = ''
+        self.stdout_hist = ''  # str (bytes in Py2, unicode in Py3)
         self.list_win = newwin(get_colpair(config, 'background'), 1, 1, 1, 1)
         self.cpos = 0
         self.do_exit = False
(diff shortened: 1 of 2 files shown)


glyph:
"bytes on python 2, unicode in python 3" is more easily referred to as "native str" :)

sebastinas:
Would be good to have test case so that we don't break that again in the future..

PR #709, opened 2017-10-10
Add Python3 classifier

Siecje:
Motivated by caniusepython3[0] reporting bpython doesn't support Python3.

```shell
$ caniusepython3 --projects bpython
```

0: https://pypi.python.org/pypi/caniusepython3


sebastinas:
If we start adding classifiers, we should also add those for Python 2.

PR #723, opened 2017-12-23
added gh-pages

extremelypenguin:
just added gh-pages

im not sure if gh-pages should be even with master or not, if i put the gh-pages folder in the wrong place please correct it


sebastinas:
The website is managed via https://github.com/bpython/bsite.

Issue #724, opened 2018-01-05
printing special characters


vaapadvaapad:
Hi Team!

Recently I've noticed that it's impossible to print string containing "\t" character using bpython and `print `function (I have 0.17 on python 3.6.3 on Fedora 27 system).
Please find below simple test case:

**python interpreter:**
```
>>> print("aaa\tbbb\tccc")
aaa	bbb	ccc
>>> 
```

**bpython:**
```
>>> print("aaa\tbbb\tccc")

>>> 
```

So I guess that '\t' prevents string from being properly displayed. For example - '\n' works correctly:

**python interpreter:**
```
>>> print("aaa\nbbb\nccc")
aaa
bbb
ccc
>>> 
```
**bpython:**
```
>>> print("aaa\nbbb\nccc")
aaa
bbb
ccc
>>>
```

Have I missed something or it can be considered as a bug? Would you please take a look at it?

Best regards and thanks a lot for your glorious work on bpython!
Andrzej

ikanobori:
Mrm, I can't reproduce this on my machine (also 3.6), can you try updating to 0.17.1 and see if the problem re-occurs. If so, I know where to look, if not I'll dive into it a bit deeper.

Issue #670, opened 2017-03-19
bpython breaks line with ANSI escape code


laike9m:
This can be easily reproduced by typing

```
print('\033[31mOK\033[0m\033[31mOK\033[0m\033[31mOK\033[0m\033[31mOK\033[0m\033[31mOK\033[0m\033[3
1mOK\033[0m\033[31mOK\033[0m\033[31mOK\033[0m\033[31mOK\033[0m\033[31mOK\033[0m\033[31mOK\033[0m\033[3
1mOK\033[0m\033[31mOK\033[0m\033[31mOK\033[0m\033[31mOK\033[0m')
```

In bpython
![image](https://cloud.githubusercontent.com/assets/2592205/24078003/758e69bc-0c9a-11e7-8dde-ce21db6f3969.png)

In Python
![image](https://cloud.githubusercontent.com/assets/2592205/24078004/831b6b2a-0c9a-11e7-8dba-035cd6a91684.png)

I guess when calculating the length of output, bpython counts all chars, but not all chars are visible, thus leading to this bug.

laike9m:
While I managed to get over the previous issue:
https://github.com/thomasballinger/curtsies/issues/97

I really can't get things to work without this being fixed. I'd appreciate if you could fix it.

laike9m:
@sebastinas @thomasballinger What do you think? Since no other REPLs has this issue, the inconsistency between them and bpython certainly brings undesired behavior to some libraries, which I think is a serious problem.

sebastinas:
Handling of escape characters happens in curtsies. So while this certainly is an issue that affects bpython, it needs to be fixed in curtsies.

thomasballinger:
Thanks to @ata2001 the second issue you mentioned https://github.com/thomasballinger/curtsies/issues/97 is fixed. Fixing length issues is possible and were it to happen via a combination of using the [wcwidth](https://github.com/jquast/wcwidth)-based width of a curtsies FmtStr. https://github.com/thomasballinger/curtsies/issues/8 is the most relevant issue for this, we should consider this bpython issue blocked on that.

laike9m:
Thanks guys. Looking forward to a new release. 😄

ata2001:
Test cases would not be bad for this, but I'd need some help with doing that.

PR #731, opened 2018-02-13
WIP - width aware output take 2

thomasballinger:
The Curtsies version referenced does not exist yet, don't merge.

Perf looks better now, a linear 10x slower. Wonder if this is good enough, I made some guesses about how to make things fast in https://github.com/bpython/curtsies/pull/109 but haven't profiled at all.

betterperf

### DIFF
--- a/bpython/curtsiesfrontend/replpainter.py
+++ b/bpython/curtsiesfrontend/replpainter.py
@@ -25,12 +25,10 @@ def display_linize(msg, columns, blank_line=False):
     """Returns lines obtained by splitting msg over multiple lines.
 
     Warning: if msg is empty, returns an empty list of lines"""
-    display_lines = ([msg[start:end]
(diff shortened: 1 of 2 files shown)


thomasballinger:
A requirement for merging this should be identifying exactly what it's fixing. I had thought it might be fixing issues like #670 but that has to do with terminal escapes - this is really only a unicode fix.

thomasballinger:
Another issue: `print('a\tsdf'*100)` crashes with this patch applied (previously how did this work?)

ikanobori:
Not entirely sure but could you, while you're in this part of the code, see if this also gets resolved: https://github.com/bpython/bpython/issues/724 with the patches?

thomasballinger:
@rybarczykj knows where this is if we work on this more, I'm closing because even if we do this it won't be with this PR.

PR #751, opened 2018-09-30
Merge send_to_stdout and send_to_stderr; fix newline handling

### LINKED ISSUE #744: Print glitch regarding newline and line-ends
The bug seems to be triggered when there is a print with `end=''` followed by a print that begins with a `\n`.

Minimal example:

    $ bpython3
    bpython version 0.17.1 on top of Python 3.6.6 /usr/bin/python3
    >>> print('foo', end=''); print('\nbar')
    foo
    
    bar
    >>>

Expected: `foo` and `bar` are in immediately adjacent lines.
Actual: There is an extraneous line between `foo` and `bar`.

For reference, observe what the standard python interpreter does:

    $ python3
    Python 3.6.6 (default, Jun 27 2018, 14:44:17) 
    [GCC 8.1.0] on linux
    Type "help", "copyright", "credits" or "license" for more information.
    >>> print('foo', end=''); print('\nbar')
    foo
    bar
    >>> 

I have the hunch that the bug probably lives in the stdio shim of bpython.  The bugtracker shows no similar bugs.

ata2001:
Fixes #744

### DIFF
--- a/bpython/curtsiesfrontend/repl.py
+++ b/bpython/curtsiesfrontend/repl.py
@@ -229,7 +229,7 @@ class ReevaluateFakeStdin(object):
             value = self.readline_results.pop(0)
         else:
             value = 'no saved input available'
-        self.repl.send_to_stdout(value)
+        self.repl.send_to_stdouterr(value)
         return value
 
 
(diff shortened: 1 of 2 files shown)


sebastinas:
`test_no_trailing_newline` fails on under Python 2.7.

ata2001:
In the meantime, I realized, that there's a lot more wrong with this, so I'm closing it for now.

PR #753, opened 2018-10-04
Travis workaround to enable CI with python3.7

ata2001:
Currently, all the python3.7 tests fail on travis because of https://github.com/travis-ci/travis-ci/issues/9069. I adjusted .travis.yml as recommended by TravisCI developer BanzaiMan in https://github.com/travis-ci/travis-ci/issues/9069#issuecomment-425720905.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ python:
     - "3.4"
     - "3.5"
     - "3.6"
-    - "3.7"
+    - "3.7-dev"
     - "3.8-dev"
     - "pypy"
     - "pypy3"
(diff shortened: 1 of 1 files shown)


ata2001:
It seems like python3.4 is not in xenial. @sebastinas should we drop 3.4 support or wait until 3.7 is supported on trusty? Perhaps we could come up with a workaround, so we can test with both versions?

sebastinas:
We should revert b214afaa44f3c7cffcc5da6b219fa1841963f152 instead. I pushed that when Travis wasn't ready.

PR #781, opened 2019-10-24
Fix AttributeError, add a test case for methods that access the history file

myrmica-habilis:
- fix `AttributeError: 'int' object has no attribute 'close'` when closing the lock file.
- add a simple test case for methods that read and/or write to the history file with the use of the `filelock.FileLock` class

### DIFF
--- a/bpython/filelock.py
+++ b/bpython/filelock.py
@@ -2,7 +2,7 @@
 
 # The MIT License
 #
-# Copyright (c) 2015-2016 Sebastian Ramacher
+# Copyright (c) 2015-2019 Sebastian Ramacher
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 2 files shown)


bpython/test/test_history.py
```diff
@@ -81,3 +84,42 @@ def test_reset(self):
...
         self.assertEqual(self.history.back(), '#999')
         self.assertEqual(self.history.forward(), '')
+
+
+class TestHistoryFileAccess(unittest.TestCase):
+    def setUp(self):
+        self.filename = 'temp_file'
+        self.encoding = getpreferredencoding()
+
+        with open(self.filename, 'w', encoding=self.encoding) as f:
```
sebastinas: This should be `io.open` to work with Python 2 too.

myrmica-habilis:
Divided into #782 and #783.

PR #787, opened 2020-01-05
Show deprecation warnings for urwid/curses/python2.

ikanobori:
@sebastinas can you take a look at mostly the texts here?


ikanobori:
Note to self: texts should be translatable.

PR #790, opened 2020-01-29
Remove desktop entry

opensource-assist:
It's very unconventional for a command-line app to have a desktop entry.
Anybody who wants to use the tool would just simply run the command from the terminal.


sebastinas:
And? The appstream metadata spec explicitly covers console applications. So, that's a NACK from me.

PR #800, opened 2020-03-26
Pinnwand api v1

supakeen:
Implement the new pinnwand API.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -10,6 +10,7 @@ General information:
 * Usage in combination with Python 2 has been deprecated. This does not mean that
   support is dropped instantly but rather that at some point in the future we will
   stop running our testcases against Python 2.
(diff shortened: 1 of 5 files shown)


CHANGELOG
```diff
@@ -10,6 +10,7 @@ General information:
 * Usage in combination with Python 2 has been deprecated. This does not mean that
   support is dropped instantly but rather that at some point in the future we will
   stop running our testcases against Python 2.
+* The new pinnwand API is used for the pastebin functionality.
```
sebastinas: Maybe add a comment that the two configuration options have been removed and that people should check the URLs in their config are sane.

supakeen: I agree I'll add that.

sebastinas:
Once that's done, should we prepare a release?

PR #812, opened 2020-06-24
Redo

allgo27:
I think I fixed it! But it does feel a little "hack-y." You previously suggested we make history a class, do you still think that's a good idea? You also suggested we move redo somewhere else, though I don't know if you had ideas of where-- do you still have thoughts on that? Thanks!

### DIFF
--- a/bpython/cli.py
+++ b/bpython/cli.py
@@ -905,7 +905,7 @@ class CLIRepl(repl.Repl):
             if n > 0:
                 self.undo(n=n)
             return ""
-
+  
         elif key in key_dispatch[config.search_key]:
             self.search()
             return ""
--- a/bpython/cli.py
+++ b/bpython/cli.py
@@ -905,7 +905,7 @@ class CLIRepl(repl.Repl):
             if n > 0:
                 self.undo(n=n)
             return ""
-  
+
         elif key in key_dispatch[config.search_key]:
             self.search()
             return ""
--- a/bpython/curtsiesfrontend/repl.py
+++ b/bpython/curtsiesfrontend/repl.py
@@ -775,13 +775,8 @@ class BaseRepl(BpythonRepl):
             self.on_tab(back=True)
         elif e in key_dispatch[self.config.undo_key]:  # ctrl-r for undo
             self.prompt_undo()
-        elif e in ["<Ctrl-g>"]:  # for redo
-            if (self.could_be_redone):
-                temp = self.could_be_redone.pop()
-                self.push(temp)
-                self.history.append(temp)
-            else:
-                self.status_bar.message("Nothing to redo.")
+        elif e in key_dispatch[self.config.redo_key]:  # ctrl-g for redo
+            self.prompt_redo()
         elif e in key_dispatch[self.config.save_key]:  # ctrl-s for save
             greenlet.greenlet(self.write2file).switch()
         elif e in key_dispatch[self.config.pastebin_key]:  # F8 for pastebin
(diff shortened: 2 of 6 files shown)


bpython/repl.py
```diff
@@ -439,7 +439,8 @@ def __init__(self, interp, config):
             duplicates=config.hist_duplicates, hist_size=config.hist_length
         )
         self.s_hist = []
-        self.history = []
+        self.history = []  # History of commands
```
thomasballinger, line 442: how about "commands executed since beginning of session," this comment isn't helping me much now.

bpython/repl.py
```diff
@@ -1009,6 +1010,10 @@ def undo(self, n=1):
 
         entries = list(self.rl_history.entries)
 
+        #Most recently undone command
+        lastEntry = self.history[-n:]
```
thomasballinger, line 1014: Could you change lastEntry -> last_entries; plural because it's a list, and snake_case

bpython/curtsiesfrontend/repl.py
```diff
@@ -860,6 +867,9 @@ def readline_kill(self, e):
 
     def on_enter(self, insert_into_history=True, reset_rl_history=True):
         # so the cursor isn't touching a paren TODO: necessary?
+        if insert_into_history:
```
thomasballinger, line 870: This if statement reads confusingly to me, I think we should:
- rename the `insert_into_history` parameter in `Repl.reevaluate()` to `new_code`; this is what it means, it's about whether the lines might be new, so they should be added to readline history.
- rename the `insert_into_history` parameter in `Repl.on_enter()` to `new_code` as well. Then this if becomes "if the line being executed is new, then we can no longer use our redo stack."
- not rename insert_into_history anywhere else

My rationale for the name is that `new_code` describes the situation/circumstance instead of the behavior we want, which is important since we're now using this flag to determine a new, different behavior.

And in reviewing this I found that the `self.reevaluate(insert_into_history=True)` in `clear_modules_and_reevaluate()` should probably be False instead, we can fix in another PR.

bpython/curtsiesfrontend/repl.py
```diff
@@ -775,6 +775,13 @@ def process_key_event(self, e):
             self.on_tab(back=True)
         elif e in key_dispatch[self.config.undo_key]:  # ctrl-r for undo
             self.prompt_undo()
+        elif e in ["<Ctrl-g>"]:  # for redo
```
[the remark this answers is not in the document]
thomasballinger: Also it'd be nice to
- follow the pattern of other keys of calling a method to keep this switch-like if/elif/elif/... structure shorter
- also for symmetry, use a tuple here instead of a list

thomasballinger:
> You previously suggested we make history a class, do you still think that's a good idea?

Could do, I don't think it's a big improvement.

> You also suggested we move redo somewhere else, though I don't know if you had ideas of where-- do you still have thoughts on that?

I'd move it to a method on this Repl class, maybe right below `prompt_undo()`, line ~1820

bpython/curtsiesfrontend/repl.py
```diff
@@ -1811,7 +1816,15 @@ def prompt_for_undo():
 
         greenlet.greenlet(prompt_for_undo).switch()
 
-    def reevaluate(self, insert_into_history=False):
+    def prompt_redo(self):
```
thomasballinger, line 1819: I think just redo, "prompt_undo" is because it may prompt you for how many lines you want to undo.

bpython/cli.py
```diff
@@ -905,7 +905,7 @@ def p_key(self, key):
             if n > 0:
                 self.undo(n=n)
             return ""
-
+
```
thomasballinger, line 908: Could you remove this extraneous change?

thomasballinger:
Ok one more thing! (I'd say on last thing but that's hard to promise)

Could you add this to the example config file, as a commented out key like undo: https://github.com/bpython/bpython/blob/d6bd2e5481bac81da4d9c23d435e7f2deb1b650c/bpython/sample-config#L69

PR #814, opened 2020-06-26
Sym two

### LINKED ISSUE #806: High CPU load when bpython is in use
Experiencing a high python process CPU load despite not running any python script. 
The issue persists as long as bpython is in use, quitting bpython resolves the high CPU load.
Running python or ipython does not cause the same issue.

79621628-aab67700-8146-11ea-927d-6e8b262a8ae0

MacOS Catalina 10.15.4
python 3.7.7: /Users/_username_/opt/anaconda3/bin/python
bpython 0.19

Apologies in advance if this issue has been addressed or the solution is trivial.
Thanks!

etiennerichart:
Use os.path.realpath to see the source of symbolic links and makes sure that we are not in a loop. Also include importtest.py and importtestfolder to test if program gets stuck in a loop.

Resolves #806


bpython/importcompletion.py
```diff
@@ -190,19 +194,21 @@ def find_modules(path):
...
-                        yield "%s.%s" % (name, subname)
+                pathReal = os.path.realpath(pathname)
+                if not pathReal in paths:
+                    paths.add(pathReal)
+                    for subname in find_modules(pathname):
+                        if subname != "__init__":
+                            yield "%s.%s" % (name, subname)
             yield name
 
-
```
thomasballinger, line 198: the convention being followed here is two lines between top-level statements. You should `pip install black` and run black on this file; that will autoformat it, dealing with these details. Or better you can [set up VSCode to run black automatically](https://code.visualstudio.com/docs/python/editing#_formatting).

importtest.py
```diff
@@ -0,0 +1,7 @@
+from bpython.importcompletion import find_modules
```
thomasballinger, line 1: If we want to keep this, you should move it to the tests folder and make it an automated test, along with these other folders needed to test this. However we probably don't want to check symlinks into version control. If you make this an automated test, I'd use functions like `os.mkdir` and `os.symlink` to build these structures in the test, run the test, then clean the folders and files up.

bpython/importcompletion.py
```diff
@@ -190,19 +194,21 @@ def find_modules(path):
...
 def find_all_modules(path=None):
     """Return a list with all modules in `path`, which should be a list of
     directory names. If path is not given, sys.path will be used."""
     if path is None:
         modules.update(try_decode(m, "ascii") for m in sys.builtin_module_names)
         path = sys.path
-
+
```
sebastinas, line 211: Extra white space change

bpython/importcompletion.py
```diff
@@ -48,6 +49,8 @@
 
 # The cached list of all known modules
 modules = set()
+#List of stored paths
```
sebastinas, line 52: Missing space after #

It would also be good to have some documentation here why the paths are stored. I'd also note here that this handles symlinks, but not mount points.

importtest.py
sebastinas: I agree, this should be part of the test suite, but `tempfile.TemporaryDirectory` and `os.symlink` should be used to setup the test.

etiennerichart:
@thomasballinger I am trying to write the test file but am having an issue with the `try:` on line 12. Also, I was wondering if the files made in the temp directory will be erased after the temp directory is closed or if they will keep taking up space.

importtest.py
```diff
@@ -0,0 +1,44 @@
...
+import os, sys, tempfile
+
+with tempfile.TemporaryDirectory() as import_test_folder:
+    try:
+        os.chdir(import_test_folder)
+    except:
+        print("Can't change the Current Working Directory")
+        sys.exit()
+
+    os.mkdir(os.path.join(os.getcwd(), "Level0")
```
thomasballinger, line 11: You're missing a closing paren on this line, that's what's causing the error on the next line

importtest.py
```diff
@@ -0,0 +1,44 @@
...
+    try:
+        os.chdir(Level0)
+        open("__init__.py", w)
+    except OSError:
+        print("Can't change the Current Working Directory")
+        sys.exit()
+
+    os.mkdir(os.path.join(os.getcwd(), "Level1")
+    try:
+        os.chdir(Level1)
```
thomasballinger, line 21: needs to be a string

importtest.py
```diff
@@ -0,0 +1,44 @@
...
+        open("__init__.py", w)
```
thomasballinger, line 22: w needs to be a string

etiennerichart:
I have almost finished turning the test class into a test file but for some reason the test is passing when it should not. In the temp directory setup I have it print out what should appear and in what order and the 'Left' file is first. In the actual test 'Left' is last and yet the test still passes.

importtest.py
```diff
@@ -0,0 +1,97 @@
+from bpython.test import unittest
+from bpython.importcompletion import find_modules
+import os, sys, tempfile
+
+with tempfile.TemporaryDirectory() as import_test_folder:
```
sebastinas, line 5: All of the above needs to happen within the test class.

etiennerichart:
I have made the changes so that importtest now works. While I was debugging I used the nosetests -s command and it seems that the temp directory is showing up in (I bolded what I think is the temp dir), '..SS.S...........S..S.......................S...SSSSS......................................................**tmpfaxxkuhx
.tmptz3ajosr**
.....S..............S.....SSS..........................S.....SS.SSSSSS.....................................................................SS....................SSSS...SS..............................S.S..........'   Is there anything else I should do to integrate the test with the rest of the test files?

sebastinas:
Please move the test file to `bpython/test`.

bpython/test/test_import_not_cyclical.py
```diff
@@ -0,0 +1,108 @@
...
+                import_test_folder.cleanup()
+                sys.exit()
+
+            os.mkdir(os.path.join(os.getcwd(), "Level0"))
+            os.mkdir(os.path.join(os.getcwd(), "Right"))
+            os.mkdir(os.path.join(os.getcwd(), "Left"))
+            try:
+                os.chdir("Level0")
+                with open("__init__.py", "w") as init_file:
+                    init_file.close()
```
sebastinas, line 22: A `pass` is sufficient here. The file will be closed when exiting this block in any case.

bpython/test/test_import_not_cyclical.py
```diff
@@ -0,0 +1,108 @@
...
+            except OSError:
+                print("Can't change the Current Working Directory")
+                sys.exit()
+
+            os.mkdir(os.path.join(os.getcwd(), "Level1"))
+            try:
+                os.chdir("Level1")
+                with open("__init__.py", "w") as init_file:
+                    init_file.close()
```
sebastinas, line 31: As above

bpython/test/test_import_not_cyclical.py
```diff
@@ -0,0 +1,108 @@
...
+            except OSError:
+                print("Can't change the Current Working Directory")
+                sys.exit()
+
+            os.mkdir(os.path.join(os.getcwd(), "Level2"))
+            try:
+                os.chdir("Level2")
+                with open("__init__.py", "w") as init_file:
+                    init_file.close()
```
sebastinas, line 40: As above

bpython/test/test_import_not_cyclical.py
```diff
@@ -0,0 +1,108 @@
...
+            try:
+                os.chdir(import_test_folder)
+            except:
+                print("Can't change the Current Working Directory")
+                sys.exit()
+
+            try:
+                os.chdir("Right")
+                with open("__init__.py", "w") as init_file:
+                    init_file.close()
```
sebastinas, line 60: As above

bpython/test/test_import_not_cyclical.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+class TestAvoidSymbolicLinks(unittest.TestCase):
+    def setUp(self):
+        with tempfile.TemporaryDirectory() as import_test_folder:
+            try:
+                os.chdir(import_test_folder)
+            except:
+                print("Can't change the Current Working Directory")
+                import_test_folder.cleanup()
```
sebastinas, line 13: This is not necessary. Once the instance of `TemporaryDirectory` is destroyed, the content in the temporary directory will be removed from the filesystem.

bpython/test/test_import_not_cyclical.py
```diff
@@ -0,0 +1,108 @@
+from bpython.test import unittest
+from bpython.importcompletion import find_modules
+import os, sys, tempfile
+
+
+class TestAvoidSymbolicLinks(unittest.TestCase):
+    def setUp(self):
+        with tempfile.TemporaryDirectory() as import_test_folder:
+            try:
+                os.chdir(import_test_folder)
```
sebastinas, line 10: Are these `chdir`s really necessary? All the paths that are calculated based on `getcwd` seem to be computable from `import_test_folder`.

bpython/test/test_import_not_cyclical.py
```diff
@@ -0,0 +1,108 @@
...
+            except:
+                print("Can't change the Current Working Directory")
+                import_test_folder.cleanup()
+                sys.exit()
```
sebastinas, line 14: I don't think that is useful. This should report as test failure instead, so I'd suggest to not catch any of these exceptions.

bpython/test/test_import_not_cyclical.py
etiennerichart: I am using the chdir to make the __init__.py files in the right folder, is there a better way?

sebastinas: Yes, you can build the paths with `os.path.join` and then pass it to `open`. For example: `open(os.path.join(import_test_folder, "Level0", "__init__.py"), "w")`. The same can be done for the `os.mkdir` calls.

bpython/test/test_import_not_cyclical.py
```diff
@@ -71,7 +71,7 @@ def test_simple_symbolic_link_loop(self):
                 self.filepaths.remove(thing)
 
 else:
-    @unittest.skip()
+    @unittest.skip("test doesn't work in python 2")
```
thomasballinger, line 74: This should be something like [skipIf](https://docs.python.org/3/library/unittest.html#unittest.skipIf) with a condition that is true for Python 2.

thomasballinger:
Now when I run this test after commenting out your fix, it still passes; so I'm worried this isn't testing what we think it is.

One problem I see is that sometimes you're using the string 'import_test_folder' when there's no folder that exists with that name. When I run your setup function (and then pause in it so I have time to examine the files) I see that the temporary directory is named `/var/folders/ty/d8c6t0c176db_ll_xh2jqn_80000gn/T/tmp5kxo1jhg` (this will be different every time) and it looks like this:
~~~
tomb ~$ tree /var/folders/ty/d8c6t0c176db_ll_xh2jqn_80000gn/T/tmp5kxo1jhg
/var/folders/ty/d8c6t0c176db_ll_xh2jqn_80000gn/T/tmp5kxo1jhg
├── Left
│   ├── __init__.py
│   └── toRight -> import_test_folder/Right
├── Level0
│   ├── Level1
│   │   ├── Level2
│   │   │   ├── Level3 -> import_test_folder/Level0/Level1
│   │   │   └── __init__.py
│   │   └── __init__.py
│   └── __init__.py
└── Right
    ├── __init__.py
    └── toLeft -> import_test_folder/Left

5 directories, 8 files
```

One this to fix would be these symlink target filenames, and then check that the test fails without your fix, but passes with it.

thomasballinger:
This fails for me locally now, but it looks close. If I sort both lists and print them, I get

~~~
actual: ['Left', 'Level0', 'Level0.Level1', 'Level0.Level1.Level2', 'Level0.Level1.Level2.Level3', 'Right', 'Right.toLeft', 'Right.toLeft.toRight']
expected: ['Left', 'Left.toRight', 'Left.toRight.toLeft', 'Level0', 'Level0.Level1', 'Level0.Level1.Level2', 'Level0.Level1.Level2.Level3', 'Right']
~~~

Is this passing for you?

It occurs to me that this could be nondeterministic since a given module can be referred to mulitple ways based on which filepath got our import discoverer there first, say if `os.listdir()` returned files in a different order we could end up with different names for the same modules. If this isn't passing for you yet either then we may not need to do this, but if it is and this is something nondeterministic we want to control we could sort the results of `os.listdir()` in the import completion code to ensure the same results each time.

etiennerichart:
I think that the problem is that when going between left and right it went right first whereas on my computer it went left first
your computer : ['Right.toLeft', 'Right.toLeft.toRight']
mine: [ 'Left.toRight', 'Left.toRight.toLeft']

thomasballinger:
Got it, I think we want to change the implementation then to sort. I think wrapping the `os.listdir()` call with sorted() like `sorted(os.listdir())` would do it.

thomasballinger:
This alternative approach works for me, import completion will traverse the filesystem in the nondeterministic order determined by os.listdir() but both orders work in the test. @sebastinas thoughts?

PR #816, opened 2020-07-01
Don't pollute history with duplicate entries during reload-and-clear-modules

allgo27:
Made small change in repl.py -- self.reevaluate(new_code=False) rather than True now.


thomasballinger:
With this change, an F6 reload-and-clear-modules or F5 automatic reload-and-clear-modules will no longer add every line of code to readline history again.

Issue #605, opened 2016-03-15
bpython cursor breaks on combining characters


aragilar:
The character à is an easy example.

thomasballinger:
Like https://github.com/bpython/bpython/issues/670 this would be fixed by using wcwidth

rybarczykj:
Resolved with PR #817! I can't find any combined characters that break cursor position anymore.

Issue #749, opened 2018-09-14
Fix handling of double-column characters


BenWiederhake:
This is probably an issue with curtsies, but until I'm sure I want to report it here first:

Steps to reproduce: Enter something like `'間'.isalnum()`, then move the cursor back (left arrow key) such that `a` is the next character, and insert an `x`.
Expected behavior: Visually it looks like we're about to insert an `x` between `is` and `alnum`.  So I would expect the line to say `isxalnum`.
Actual behavior: It now says `'間'.isaxlnum()`, with the next character seemingly being the `x`.

Try this out with `y` instead of `間` to see the expected behavior.

I may have broken this in the patch to #703, I'm not sure.

The pun ("間" meaning "between") is entirely coincidental, but I like it, and I'm keeping it.

rybarczykj:
Resolved with PR #817! And a current PR is in progress to fix line wrapping with double-column chars.

sebastinas:
Fixed thanks to @rybarczykj

PR #817, opened 2020-07-01
Corrected cursor positioning on inputs with full-width (2 column) characters

rybarczykj:
This addresses issue #749. It should now be fixed for normal input as well as stdin input.

Before:
widetextbefore

After:
widetextafter

Also, because the issue with width aware text wrapping still exists (#731), it gets a little weird when you exceed one line. Still, I think the cursor is in the correct spot. In the below example the un-wrapped part is '６７８９０'. The cursor seems right despite the text not wrapping.
widetextwrap

### DIFF
--- a/bpython/curtsiesfrontend/repl.py
+++ b/bpython/curtsiesfrontend/repl.py
@@ -19,6 +19,8 @@ from pygments import format as pygformat
 from bpython._py3compat import PythonLexer
 from pygments.formatters import TerminalFormatter
 
+from wcwidth import wcswidth
+
 import blessings
 
 import curtsies
@@ -1556,7 +1558,8 @@ class BaseRepl(BpythonRepl):
 
         if self.stdin.has_focus:
             cursor_row, cursor_column = divmod(
-                len(self.current_stdouterr_line) + self.stdin.cursor_offset,
+                len(self.current_stdouterr_line)
+                + wcswidth(self.stdin.current_line[: self.stdin.cursor_offset]),
                 width,
             )
             assert cursor_column >= 0, cursor_column
@@ -1577,9 +1580,9 @@ class BaseRepl(BpythonRepl):
         else:
             cursor_row, cursor_column = divmod(
                 (
-                    len(self.current_cursor_line_without_suggestion)
-                    - len(self.current_line)
-                    + self.cursor_offset
+                    self.current_cursor_line_without_suggestion.width
+                    - wcswidth(self.current_line)
+                    + wcswidth(self.current_line[: self.cursor_offset]) 
                 ),
                 width,
             )
--- a/bpython/curtsiesfrontend/repl.py
+++ b/bpython/curtsiesfrontend/repl.py
@@ -1361,7 +1361,11 @@ class BaseRepl(BpythonRepl):
 
     @property
     def current_cursor_line_without_suggestion(self):
-        """Current line, either output/input or Python prompt + code"""
+        """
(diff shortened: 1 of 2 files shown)


thomasballinger:
This is great, nice.

Could you check whether this works + whether the .width method exists in the lowest required version of curtsies, as defined in the bpython setup.py?

This might be good as is, but some other pieces to look at:
* when I paste "`ｈｅ ａｂｏｖｅ， ｔｈｅｒｅ ａｒｅ`" into the save dialog, I get greater-than signs. What's up with that?
* the cursor in input("ｈｅ ａｂｏｖｅ， ｔｈｅｒｅ ａｒｅ ") isn't aligned yet
* wrapping, as you've pointed out
* F7 works, great!
* Are there any examples you can find that are a the wrong width, as per https://jeffquast.com/post/terminal_wcwidth_solution/? If we we should open an issue and you might look at this too.

thomasballinger:
Another thing, it'd be great to close https://github.com/bpython/bpython/pull/731 once this or another PR supersedes it for fixing wrapping

thomasballinger:
Could you confirm that this fixes https://github.com/bpython/bpython/issues/605 too?

thomasballinger:
and https://github.com/bpython/bpython/issues/670!

rybarczykj:
> when I paste "ｈｅ ａｂｏｖｅ， ｔｈｅｒｅ ａｒｅ" into the save dialog, I get greater-than signs. What's up with that?

Oh yeah, that's weird. I think I'll leave that for now though. It seems outside of scope.

>the cursor in `input("ｈｅ ａｂｏｖｅ， ｔｈｅｒｅ ａｒｅ ")` isn't aligned yet

Should be fixed with commit 9e380


> wrapping

Yep, inputs with wide chars now work fine until you exceed one line. I'll look at this next.

> #605 

This seems fixed now, at least for all non-emoji combined characters I could find. Some emojis that combine characters do actually break it, but that's obviously not very urgent. For reference the emoji thing is a [known wcwidth issue](https://github.com/jquast/wcwidth/issues/39).

bpython/curtsiesfrontend/repl.py
```diff
@@ -1577,9 +1580,9 @@ def move_screen_up(current_line_start_row):
         else:
             cursor_row, cursor_column = divmod(
                 (
-                    len(self.current_cursor_line_without_suggestion)
-                    - len(self.current_line)
-                    + self.cursor_offset
+                    self.current_cursor_line_without_suggestion.width
+                    - wcswidth(self.current_line)
```
thomasballinger, line 1584: What's the distinction here between the wcswidth and .width, are some of these FmtStrs and some of them plain strs? (and if it's not clear, could you add comments to the definitions or initialization lines?)

thomasballinger:
Sorry about the delay here. This looks good, only thing left is to clarify why we're using both techniques instead of just one or the other, since dropping either (removing wcswidth when we haven't added it to setup.py, or eliminating the need for the Curtsies upgrade) would be nice. Also, assuming we do keep the wcwidth library, it should be added to the explicit setup.py dependencies.

rybarczykj:
Oh yep. I had used both because some of those objects were FmtStr and some str, but I can definitely make it only use wcswidth() to avoid the curtsies upgrade for now, because `FmtStr.width` is essentially a wrapper for `wcswidth(FmtStr.s)`. I'll do that and update the setup.py

PR #818, opened 2020-07-09
From <modules> import <> completion

etiennerichart:
Trying to create background import completion for unimported modules when calling from <modules> import ... . So that ... will autocomplete. Am using numpy to test code and so have numpy not imported when start loading.

### DIFF
--- a/bpython/autocomplete.py
+++ b/bpython/autocomplete.py
@@ -251,6 +251,9 @@ class ImportCompletion(BaseCompletionType):
     def format(self, word):
         return after_last_dot(word)
 
+    def try_to_complete(self, module):
+        return importcompletion.try_to_import(module)
+
 
 class FilenameCompletion(BaseCompletionType):
     def __init__(self, mode=SIMPLE):
--- a/bpython/autocomplete.py
+++ b/bpython/autocomplete.py
@@ -251,8 +251,11 @@ class ImportCompletion(BaseCompletionType):
     def format(self, word):
         return after_last_dot(word)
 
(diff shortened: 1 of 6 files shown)


etiennerichart:
For some reason the code in module_matches is not calling find_all_modules successfully.

thomasballinger:
I'm hoping for module attribute completion, so we'll need the real module objects, not just what we can get from the filesystem with import completion. I think on tab we should actually import the module, so we can provide completions.

I think this might just work if we import the module on pressing tab. The `on_tab()` method here https://github.com/bpython/bpython/blob/214e4ca63a9643a4a3f72081542fe187a0587ad8/bpython/curtsiesfrontend/repl.py#L887 would be the place to start with this, if it looks like the user was writing `from foo import <tab>` then go ahead an import foo.

etiennerichart:
I believe it works. Try running bpython and then writing:
from numpy import a #You should only be getting anything at the moment
from numpy import a #Press the tab key and then you will get the attribute array. 

You can check that it is the numpy attribute array by using it like this:
num = array([2, 3])
num2 = num + 2 --> [4, 5]

thomasballinger:
This works for me, a few thoughts
- let's check to see if we really need to try to import the module before displayign the message and trying
- I'd like the `Importing module numpy` banner to disappear as soon as the import has finished, I assumed that bpython was frozen importing it while it was still up
- this might be a big change, but it would be nice to show completion options on tab, even if it's an blank word we're completing

Issue #727, opened 2018-01-18
End of line cut off after saving tmp file


waltsims:
One line with a long data path was cut off after I edited the history file and saved it back to the tmp file.  Is this a known bug that I missed in the issues? Maybe it has something to do with line wrapping?

ikanobori:
Can you give me some more direct instructions on how to reproduce this so I can take a look?

thomasballinger:
Repro:
* start bpython in a narrow terminal
* `>>> s = 'very long string literal that is going to be cut off if you make your terminal very narrow'`
* F7
* edit the string
* save
* observe error on reevaluation

ikanobori:
@thomasballinger I know you were taking a look as per IRC, did you make any progress on this?

sebastinas:
FWIW, the issue is in `bpython.curtsiesfrontend.BaseRepl.send_session_to_external_editor`. It reconstructs the user input from `self.getstdout()` instead of the history. Thus long lines, that are a single input line, but require multiple lines to be displayed, contain extra line breaks.

PR #821, opened 2020-07-14
Fix long lines sent to editor

rybarczykj:
This is a working fix to #727, where long lines sent to the editor (usually with F7) are cut off and can't be edited correctly.

From what I can tell, the solution seems to work fine, but since this is the most code I've written at once for bpython, I want to make sure the structure / conventions / methods are valid. So I'll start it as a draft.

### DIFF
--- a/bpython/curtsiesfrontend/repl.py
+++ b/bpython/curtsiesfrontend/repl.py
@@ -31,6 +31,7 @@ from curtsies import events
 
 import bpython
 from bpython.repl import Repl as BpythonRepl, SourceNotFound
+from bpython.repl import LineTypeTranslator as LineType
 from bpython.config import (
     Struct,
     loadini,
@@ -396,7 +397,6 @@ class BaseRepl(BpythonRepl):
         # current line of output - stdout and stdin go here
         self.current_stdouterr_line = ""
 
-
         # this is every line that's been displayed (input and output)
         # as with formatting applied. Logical lines that exceeded the terminal width
         # at the time of output are split across multiple entries in this list.
@@ -408,8 +408,9 @@ class BaseRepl(BpythonRepl):
         # This is every logical line that's been displayed, both input and output. 
         # Like self.history, lines are unwrapped, uncolored, and without prompt.
         # Entries are tuples, where
-        #   the first element a string of the line
-        #   the second element is one of 2 strings: "input" or "output".
+        #   - the first element the line (string, not fmtsr)
+        #   - the second element is one of 2 global constants: "input" or "output"
+        #     (use LineType.INPUT or LineType.OUTPUT to avoid typing these strings)
         self.all_logical_lines = []
 
         # formatted version of lines in the buffer kept around so we can
@@ -881,9 +882,8 @@ class BaseRepl(BpythonRepl):
             self.rl_history.reset()
 
         self.history.append(self.current_line)
-        self.all_logical_lines.append((self.current_line, "input"))
+        self.all_logical_lines.append((self.current_line, LineType.INPUT))
         self.push(self.current_line, insert_into_history=new_code)
-
 
     def on_tab(self, back=False):
         """Do something on tab key
@@ -1016,7 +1016,7 @@ class BaseRepl(BpythonRepl):
         """
         for_editor = EDIT_SESSION_HEADER
         for_editor += "\n".join(
-            line[0] if line[1] == "input"
+            line[0] if line[1] == INPUT
             else "### " + line[0] 
             for line in self.all_logical_lines
         )
@@ -1189,9 +1189,7 @@ class BaseRepl(BpythonRepl):
         if c:
             logger.debug("finished - buffer cleared")
             self.cursor_offset = 0
-
             self.display_lines.extend(self.display_buffer_lines)
-            
             self.display_buffer = []
             self.buffer = []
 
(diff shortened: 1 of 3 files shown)


bpython/curtsiesfrontend/repl.py
```diff
@@ -395,9 +396,9 @@ def __init__(
         # current line of output - stdout and stdin go here
         self.current_stdouterr_line = ""
 
-        # lines separated whenever logical line
-        # length goes over what the terminal width
-        # was at the time of original output
+        # this is every line that's been displayed (input and output).
+        #   seperated logically, but if a line exceeded the terminal width
+        #   at the time of output, it is split into multiple entries
```
thomasballinger, line 401:
```suggestion
        # this is every line that's been displayed (input and output)
        # as with formatting applied. Logical lines exceeded the terminal width
        # at the time of output are split across multiple entries in this list.
```

bpython/curtsiesfrontend/repl.py
```diff
@@ -411,7 +412,8 @@ def __init__(
         # because there wasn't room to display everything
         self.scroll_offset = 0
 
-        # from the left, 0 means first char
+        # cursor position relative to start of current_line
+        #   0 means first char
```
thomasballinger, line 416:
```suggestion
        # cursor position relative to start of current_line, 0 is first char
```

bpython/curtsiesfrontend/repl.py
```diff
@@ -994,15 +996,58 @@ def send_current_block_to_external_editor(self, filename=None):
         self.cursor_offset = len(self.current_line)
 
     def send_session_to_external_editor(self, filename=None):
+        """
+        Sends entire bpython session to external editor to be edited. Usually called with F7.
```
thomasballinger, line 1000:
```suggestion
        Sends entire bpython session to external editor to be edited. Usually bound to F7. 
```
to me "called with" means the arguments passed in

bpython/curtsiesfrontend/repl.py
```diff
@@ -994,15 +996,58 @@ def send_current_block_to_external_editor(self, filename=None):
         self.cursor_offset = len(self.current_line)
 
     def send_session_to_external_editor(self, filename=None):
+        """
+        Sends entire bpython session to external editor to be edited. Usually called with F7. 
+
+        Loops through display_lines and writes lines to editor. Whenever a line starts with ps1 or ps2, 
+        we actually pop and write from history instead of display_lines. This is because display_lines 
+        cuts off long lines but self.history preserves them.
```
thomasballinger, line 1004: ooo tricky!

bpython/curtsiesfrontend/repl.py
```diff
@@ -1903,6 +1948,11 @@ def initialize_interp(self):
         del self.coderunner.interp.locals["_Helper"]
 
     def getstdout(self):
```
thomasballinger, line 1950: Heads up that this method is also called by write2file(), copy2clipboard(), and pastebin() in bpython/repl.py so we may or may not want to change its behavior (depends on whether the changes make sense for those methods too)

rybarczykj: Oh whoops, I accidentally left that outdated TODO in there. Yeah I agree, we shouldn't need to change this method.

rybarczykj:
Ok as Tom and I discussed (outside of this thread), I've added a new list, `self.all_logical_lines` which keeps track of all input and output lines without wrapping or formatting. The data structure is a list of tuples, where the first element is a string of the line and the second element is one of 2 strings: "input" or "output". From all the tests and checks I ran, I'm pretty sure I've added it in all the correct spots so that it updates and undoes correctly.

This makes it very easy to write to editor without having to worry about wrapping.

Also, as we talked about, possible future extensions of this would be making `self.history` dependent on this list, to save space (`self.history` is always equal to all "input" lines in `self.all_logical_line`). This list could be useful to use with copying and pasting code blocks, where wrapping is so commonly an issue.

bpython/curtsiesfrontend/repl.py
```diff
@@ -1870,8 +1868,8 @@ def prompt_for_undo():
     def redo(self):
         if (self.redo_stack):
             temp = self.redo_stack.pop()
-            self.push(temp)
             self.history.append(temp)
+            self.push(temp)
```
thomasballinger, line 1872: What was the significance of this, why is it important to push append to history first?

bpython/curtsiesfrontend/repl.py
```diff
@@ -1847,12 +1837,20 @@ def take_back_buffer_line(self):
             self.cursor_offset = len(self.current_line)
         self.display_buffer.pop()
         self.buffer.pop()
-        self.history.pop()
+        undone_input = self.history.pop()
+
+        # Pop all outputs leading up to the undone input
```
thomasballinger, line 1842: We shouldn't need this for `take_back_buffer_line()`, the buffer is the current (as yet unexecuted) possibly-multiline input statement, so removing a line will not affect output

bpython/curtsiesfrontend/repl.py
rybarczykj: Oh, right. It makes no difference the order of those two.

But, I meant to also append to all_logical_lines right there, which _does_ have to be before self.push(temp). Rearranging the order of self.history would keep those three steps in the same order as is in on_enter.

Adding that commit now to add all_logical_lines there.

thomasballinger: cool, seems fine to reorder for consistency

bpython/curtsiesfrontend/repl.py
```diff
@@ -408,13 +409,16 @@ def __init__(
...
         # Entries are tuples, where
         #   - the first element the line (string, not fmtsr)
-        #   - the second element is one of 2 global constants: INPUT or OUTPUT 
-        #     indicating the type of line
+        #   - the second element is one of 2 global constants: "input" or "output",
         self.all_logical_lines = []
+        
+        # This is to use "input" or "output" for all_logical_lines tuples without actually 
+        # having to know the string.
+        class lineTypeTranslator(Enum):
```
thomasballinger, line 417: I'd move this out to the global scope so that other code can access it without needing to use self.lifeTypeTranslator, and use TitleCase instead of camelCase

rybarczykj: Got it. I added it as a class in bpython.repl, and now curtsiesfrontend.repl imports it.

PR #828, opened 2020-07-22
Fix issue returning from editor

rybarczykj:
Small fix on a issue returning from editor which occurs after some changes I made to the commenting style on my last commit.

Side note: is it fine that some non-input lines in the editor session start with "# OUT:" and others with "### "? I think it's helpful for clarity, but I recognize that having the same prefix for all line is consistent.

### DIFF
--- a/bpython/curtsiesfrontend/repl.py
+++ b/bpython/curtsiesfrontend/repl.py
@@ -1026,7 +1026,7 @@ class BaseRepl(BpythonRepl):
             current_line = lines[-1][4:]
         else:
             current_line = ""
-        from_editor = [line for line in lines if line[:6] != "# OUT:"]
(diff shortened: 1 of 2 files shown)


thomasballinger:
What's the difference between these different types of non-input lines?

rybarczykj:
"#OUT: " for output lines
"### " for current line and EDIT_SESSION_HEADER

The previous version of get_session_formatted_for_file() preceded lines with "#OUT: " but when I changed it to include the current line I didn't think it made sense to say the current line was #OUT.

Should I just make all lines "### "?

thomasballinger:
Oh that makes sense, I like how you had it - distinguishing output from text we've added seems smart.

rybarczykj:
This should be ready, and just to be clear about what this specific PR does: I fixed this issue where it displays the commented out lines when you return to REPL.

Screen Shot 2020-07-27 at 2 44 14 PM

PR #831, opened 2020-07-24
Remove Enum to maintain Python 2 support

thomasballinger:
FYI @rybarczykj


sebastinas:
For Python 2.7, [enum34](https://pypi.org/project/enum34/) could be used instead.

thomasballinger:
That's fair, this is just for one class so I'll leave it for now.

PR #834, opened 2020-07-30
Config editors

etiennerichart:
Added information on how to use VS Code and Atom as your default external editor. Only tested on Mac.

### DIFF
--- a/bpython/sample-config
+++ b/bpython/sample-config
@@ -36,7 +36,7 @@
 # color_scheme = default
 
 # External editor to use for editing the current line, block, or full history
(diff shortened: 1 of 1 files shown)


bpython/sample-config
```diff
@@ -36,6 +36,10 @@
 # color_scheme = default
 
 # External editor to use for editing the current line, block, or full history
+# Options: vi (vim)
```
sebastinas, line 39: I'd call that examples. Options would suggest that these are the only supported values.

sebastinas:
Thanks, rebased and merged.

PR #836, opened 2020-07-31
ImportFinder: Support find_spec interface

### LINKED ISSUE #835: bpython shell cannot import python namespace packages
In Python 3.3, python namespace packages were introduced [0] - namespace package is a virtual package whose contents can be distributed in various places along Python's PATH. TL;DR 

Currently importing a namespace package in bpython fails with:
```
>>> import my_namespace
  File "<input>", line 1, in <module>
    import my_namespace
ModuleNotFoundError: No module named 'my_namespace'
```

whereas the identical python distribution, just running native shell, produces:

```
>>> import my_namespace
>>> my_namespace
<module 'my_namespace' (namespace)> 
```

[0] https://www.python.org/dev/peps/pep-0420/

tbabej:
The absence of `ImportFinder` wrapping the `find_spec` method (introduced in Python 3.4), which is a successor to now deprecated `find_module` [0] cause one of the finder classes to not find the namespace package, as it is only detectable using `find_spec`

Implements backwards-compatible logic to support `find_spec` if present in the Finder being wrapped.

Closes #835.

[0] https://docs.python.org/3/library/importlib.html#importlib.abc.MetaPathFinder.find_spec

### DIFF
--- a/bpython/curtsiesfrontend/repl.py
+++ b/bpython/curtsiesfrontend/repl.py
@@ -287,6 +287,18 @@ class ImportFinder(object):
 
         return None
 
+    def find_spec(self, fullname, path, target=None):
+        for finder in self.old_meta_path:
+            # Consider the finder only if it implements find_spec
+            if not getattr(finder, "find_spec", None):
+                continue
+            # Attempt to find the spec
+            spec = finder.find_spec(fullname, path, target)
+            if spec is not None:
+                # Patch the loader to enable reloading
+                spec.__loader__ = ImportLoader(self.watcher, spec.__loader__)
+                return spec
+
     def find_module(self, fullname, path=None):
         for finder in self.old_meta_path:
             loader = finder.find_module(fullname, path)


sebastinas:
@tbabej Looks like that didn't cover all of it:

```
bpython version 0.20.dev10 on top of Python 3.8.5 /usr/bin/python3             
  File "<input>", line 1, in <module>                                          
    from bpython.curtsiesfrontend._internal import _Helper                  
  File "<frozen importlib._bootstrap>", line 991, in _find_and_load
  File "<frozen importlib._bootstrap>", line 971, in _find_and_load_unlocked                                                                                   
  File "<frozen importlib._bootstrap>", line 914, in _find_spec  
  File "/home/sebastian/development/bpython/bpython/curtsiesfrontend/repl.py", line 299, in find_spec
    spec.__loader__ = ImportLoader(self.watcher, spec.__loader__)              
AttributeError: 'ModuleSpec' object has no attribute '__loader__'              
  File "<input>", line 1, in <module>                                          
    help = _Helper(_repl)            
NameError: name '_Helper' is not defined   
```

I've now changed it in afc28c3fb20a23c497bca11a0a6afb1b22bd6c97 to only replace `__loader__` if defined and not `None`. Do you think that this makes sense?

PR #839, opened 2020-08-05
Add width awareness to fix wrapping of strings with full width chars

rybarczykj:
This is a redux of #731 which was put on hold until a need for width awareness was found. I'd say this is a decent justification for bringing it back 

**Wrapping strings with fullwidth chars** 
Here's the same string being typed before and after width awareness:
image

image

**New method to fix cursor position:**
When we put a 2-column character at the end of the line where there's only 1 column of room, curtsies adds a padding character (space) so we can kick it down to the line below. Therefore, in order to get the right cursor position, I had to add a method that determines how many padding chars there are (`number_of_padding_chars_on_current_cursor_line`). 

Note on this method: Right now, to determine the amount of padding, it calls `display_linize`, doubling the amount of times we call that function. It's definitely possible to keep track of this as a variable in the repl class instead, or to do some changes other to ensure less redundant work is done. Feedback appreciated.

**Runtime**
One of the barriers stopping us from implementing this before was runtime . I did not find any noticeable difference when pasting, printing, or moving my cursor around large strings.

### DIFF
--- a/bpython/curtsiesfrontend/repl.py
+++ b/bpython/curtsiesfrontend/repl.py
@@ -402,7 +402,7 @@ class BaseRepl(BpythonRepl):
         # so we're just using the same object
         self.interact = self.status_bar
 
-        # line currently being edited, without ps1 (usually '>>> ')
+        # logical line currently being edited, without ps1 (usually '>>> ')
         self._current_line = ""
 
         # current line of output - stdout and stdin go here
@@ -744,8 +744,9 @@ class BaseRepl(BpythonRepl):
             )
             and self.config.curtsies_right_arrow_completion
             and self.cursor_offset == len(self.current_line)
+            # if at end of current line and user presses RIGHT (to autocomplete)
         ):
-
+            # then autocomplete
             self.current_line += self.current_suggestion
             self.cursor_offset = len(self.current_line)
         elif e in ("<UP>",) + key_dispatch[self.config.up_one_line_key]:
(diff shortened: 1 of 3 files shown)


bpython/curtsiesfrontend/repl.py
```diff
@@ -1435,6 +1436,25 @@ def current_output_line(self, value):
...
 
+    def number_of_padding_chars_on_current_cursor_line(self):
+        """ To avoid cutting off two-column characters at the end of lines where
+        there's only one column left, curtsies adds a padding char (u' ').
+        It's important to know about these for cursor positioning.
+
+        Should return zero unless there are fullwidth characters. """
+        full_line = self.current_cursor_line_without_suggestion
+        line_with_padding = "".join(
+            [
```
sebastinas, line 1447: I don't think this needs to be a list. `join` works with a generator just fine.

sebastinas:
Could you please rebase this PR?

rybarczykj:
Just noticed a bug with this; don't accept yet.

In display_linize, before calling `width_aware_splitlines`, we convert the msg to a format string: `msg = fmtstr(msg)` . However, this causes and error if msg is _already_ a format string, which can happen if you try to display your own format strings by importing curtsies 

I could hard code this case to check if it's already a format string, but maybe it's better to adjust the initializer of fmtstr() to do the work in curtsies. @thomasballinger thoughts?

rybarczykj:
That bug I mentioned ^^ is fixed now, as of that last push. Now, if wcswidth throws an error, we catch it and use the old display_linize.

I also added some relevant tests.

PR #840, opened 2020-08-06
Feature: Add open close parentheses for attribute suggestions that are methods

rybarczykj:
TLDR:
- Parenthesis on attribute suggestions so it's clear with attributes are methods
- Slight modification to how parentheses are autofilled

Sorry for the bad quality gifs. I should stop using GIPHY for this.

**Native function completion**
Current behavior: open parenthesis in suggestion and autofill. You have to type something after the paren before you can see the method signature.
now
My proposal: open AND close parens in suggestion, no parens in autofill. Now typing the open paren shows the method signature.
new
This might mess with some people's muscle memory since it changes what's autofilled.

**Attributute completion**
Current behavior: no parens in suggestion or autofill. You can't tell what is a method or otherwise.
old
My proposal: open close parens in suggestion, no parens in autofill. Very clear which attrs are methods.
old
Since this only changes the infobox, it shouldn't mess with anyone's muscle memory.

If this seems worth implementing, I'll need to fix some tests. Right now, tests expect matches that look like "math.cos" or "any(", but they'll be getting matches that look like "math.cos()" and "any()". 

Also I'm not sure if I implemented this the best way. The way I did it, the actual match string is altered to have parens. Maybe it's better to leave the match as-is, but add parenthesis only on the front end. This would require some passing around of information. Feedback appreciated.

### DIFF
--- a/bpython/autocomplete.py
+++ b/bpython/autocomplete.py
@@ -119,6 +119,8 @@ else:
 
 
 def after_last_dot(name):
+    """matches are stored as 'math.cos', 'math.sin', etc. This function returns
+    just 'cos' or 'sin' """
     return name.rstrip(".").rsplit(".")[-1]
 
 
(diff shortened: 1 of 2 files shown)


thomasballinger:
I personally like this, and I think at least the second part, attribute completion changes, should be added.

PR #844, opened 2020-08-18
Advanced import completion

etiennerichart:
when typed: from <module> import <> and tab is pressed will background import the module to get better autocomplete options. Note: If nothing is yet typed after the import the autocomplete will not show any options.

### DIFF
--- a/bpython/curtsiesfrontend/repl.py
+++ b/bpython/curtsiesfrontend/repl.py
@@ -903,7 +903,6 @@ class BaseRepl(BpythonRepl):
         if "from" in line and "import" in line:
             # returns <module> name if the line contains: "from <module> import"
             module_name = from_import_tab(self.current_line)
-            print(module_name)
             if module_name:
                 if module_name not in sys.modules:
                     self.status_bar.push_permanent_message(
--- a/bpython/curtsiesfrontend/repl.py
+++ b/bpython/curtsiesfrontend/repl.py
@@ -37,6 +37,7 @@ from bpython.formatter import BPythonFormatter
 from bpython import autocomplete
 from bpython.translations import _
 from bpython.pager import get_pager_command
+from bpython.line import from_import_tab
 
 from bpython.curtsiesfrontend import replpainter as paint
 from bpython.curtsiesfrontend import sitefix
@@ -887,6 +888,15 @@ class BaseRepl(BpythonRepl):
             for unused in range(to_add):
                 self.add_normal_character(" ")
             return
+        
+        # if line is `from a import` and tab then import module and attr
+        line = self.current_line.split()
+        module_name = ""
+        if "from" in line and "import" in line:
+            module_name = from_import_tab(self.current_line)
+        if module_name:
+            self.status_bar.message("Importing module %s" % module_name)
+            self.completers[1].try_to_complete(module_name)
 
         # run complete() if we don't already have matches
         if len(self.matches_iter.matches) == 0:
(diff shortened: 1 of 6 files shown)


sebastinas:
I'm not sure what this PR tries to achieve. What's the behavior before and after this change?

thomasballinger:
The goal of this PR is that when typing out an import for a module (numpy for example)
`>>> from numpy import a<tab>` should display all submodules and module attributes. Hitting tab might cause numpy to be imported too.

I don't know the status of this PR, it's been on my list to review for a bit; I worked with @etiennerichart on it but haven't been paying attention for a couple months now.

sebastinas:
Ah, yes, that's indeed useful.

bpython/curtsiesfrontend/repl.py
```diff
@@ -897,6 +906,20 @@ def only_whitespace_left_of_cursor():
                 self.add_normal_character(" ")
             return
 
+        # 5. if line is `from a import` and tab then import module and attr
+        line = self.current_line.split()
+        module_name = ""
+        if "from" in line and "import" in line:
+            # returns <module> name if the line contains: "from <module> import"
+            module_name = from_import_tab(self.current_line)
+            print(module_name)
```
sebastinas, line 915: This is a left over debug print, I guess.

sebastinas:
We just removed 0.20 to fix compatibility with Python 3.9 and now dropped support for Python 2 in master. The latter touched almost the whole code base. Could you please rebase this PR on top of current master? Thanks

bpython/line.py
```diff
@@ -120,6 +120,19 @@ def current_object_attribute(cursor_offset, line):
     return None
 
 
+from_import_tab_re = LazyReCompile(r"from ([\w0-9_.]*)\s+import")
+
+
+def from_import_tab(line):
+    matches = from_import_tab_re.finditer(line)
+    try:
+        module = matches.__next__()
```
sebastinas, line 129: This should be `next(matches)`, right?

bpython/importcompletion.py
```diff
@@ -99,6 +100,19 @@ def module_attr_matches(name):
     return attr_matches(name, prefix="", only_modules=True)
 
 
+def try_to_import(module_name):
```
sebastinas, line 103: Is this really necessary? Shouldn't those modules already be found by our normal module processing routines?

etiennerichart:
I have not been able to test functionality yet since I'm having trouble starting bpython on my PC. I'm getting the following error:
`File "c:\users\etien\appdata\local\programs\python\python38-32\lib\site-packages\blessings\__init__.py", line 5, in <module>
    import curses
  File "c:\users\etien\appdata\local\programs\python\python38-32\lib\curses\__init__.py", line 13, in <module>
    from _curses import *
ModuleNotFoundError: No module named '_curses'`
If someone wants to test it for me, just try:
`from <module> import <at least one letter>` seeing if there are autocomplete suggestions, then pressing tab to see if you get more suggestions. I usually test on using numpy module but you may have to pip install it.

sebastinas:
As far as I can tell, the PR works as expected. I'm just not sure if we need the additional module discovery code.

sebastinas:
Okay, I took a look at this PR again. I don't think we should merge it in its current state. `from x import y` auto completion is already in handled in `bpython.importcompletion` and `bpython.autocomplete`. The issue that this PR tries to solve is auto completion when our import completion hasn't processed `x`. What we could discuss if we can fix the import completion code to process `x` in `bpython.importcompletion.complete` if it is not yet known.

But I don't think we should add a curtsies specific workaround that breaks abstraction.

PR #851, opened 2020-10-02
added content in __init__.py

samsy12:


sebastinas:
But why?

bpython/cli.py
```diff
@@ -672,7 +672,7 @@ def get_key(self):
     def get_line(self):
         """Get a line of text and return it
         This function initialises an empty string and gets the
-        curses cursor position on the screen and stores it
+        cursor position on the screen and stores it
```
sebastinas, line 675: That's actually correct.

sebastinas (changes requested):

sebastinas:
I don't think that is useful.

samsy12:
Can you please say once more what was the issue with my pr?
And can you please not mark it as invalid/spam? Please, it's kind of
inportant.
No need to merge it, just if you could please remove the invalid/spsm
label, it would be great.

sebastinas:
That doesn't change the fact that the spelling fix is wrong and for the other the open question of "why?" is still unanswered.

samsy12:
Alright, give me some time, I'll review the code and get back to you.
Meanwhile, can you please remove the invalid label on this pr?
I have some limitations on the amount of data I can consume per day and
with all classes being online, there is only so much other stuff I can do
online. I hope you understand.

PR #854, opened 2020-10-11
Fix invalid escape sequences.

thomasballinger:
Two of these come from 0259cf8 and I'm not sure what the original intent was of checking for a curses key code like "\S" or "\T". I'm guessing they should be ^S and ^T, but ^S is usually already used for save. I'm going to remove these.


sebastinas:
LGTM

PR #856, opened 2020-10-11
Restore original save behavior for cli and urwid

### LINKED ISSUE #855: Saving broken in bpython-curses
Unfortunately https://github.com/bpython/bpython/pull/821 seems to have broken saving in bpython-curses because write2file() now relies on the Repl child class attribute all_logical_lines, which cli.py does not implement.

thomasballinger:
Resolves #855


sebastinas:
LGTM

PR #862, opened 2020-11-10
upgrading curtsies version to include changes from PR 111 to fix typing bug

kelseylb:
https://github.com/bpython/curtsies/pull/111 fixed an issue that causes a crash in mypy, but bpython isn't guaranteed to use a version of curtsies that includes this fix. this diff upgrades the bpython dependency on curtsies to 0.3.3, the latest release.

i couldn't find a contributing.md or similar to show how to build and test the diff, or with any other guidelines for contributors, so if there are additional steps i should take before this is shippable, please let me know!

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 Pygments
-curtsies >=0.3.2
+curtsies >=0.3.3
 greenlet
 requests
 setuptools
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 Pygments
-curtsies >=0.3.0
+curtsies >=0.3.2
 greenlet
 requests
 setuptools


kelseylb:
thanks for the merge, @sebastinas! what should I expect in terms of process to see this released?

sebastinas:
Unless some major issue affecting 0.20.1 pops up, we haven't planned any release until the beginning of 2021. But if we need to do a 0.20.2, it will be included. Otherwise it will be part of 0.21.

kelseylb:
when do you expect to release 0.21? this bug is blocking our adoption of mypy, and we need to either fix it or migrate off of bpython to unblock that effort.

sebastinas:
I'm not sure I'm following. The current requirments does not prevent upgrading curtsies to 0.3.3. @supakeen Any opinion on that?

supakeen:
I currently get curtsies 0.3.4 on a fresh install:

```
 € python3.8 -m venv venv
...
€ venv/bin/pip install bpython
...
€ venv/bin/pip freeze | grep curtsies
curtsies==0.3.4
```

Can you tell me how you're receiving the older version of `curtsies`, if there's a way to get that then a 0.20.2 is warranted.

PR #863, opened 2020-11-11
Docs state wrong default config file directory

niloct:
Fixing correct default directory for bpython config file (especially useful for color customization).


niloct:
Seems like there was a regression ?

https://docs.bpython-interpreter.org/en/latest/configuration.html#general

sebastinas:
No, the docs just haven't been rebuilt since the PR was merged.

PR #909, opened 2021-08-22
Fixed bug with sys.stdin.readline()

linux-admin0001:
image
I fixed this bug:
image


sebastinas:
I fear that this just hides an issue. `self.current_stdouterr_line` has the wrong time in this case and that should be fixed.

PR #920, opened 2021-09-27
Changed dict key-matching regex to capture any valid dict key

### LINKED ISSUE #917: Tab completion for dict key causes crash
This issue is not reproducible with several different `dict`s that I've tried, but it can be reproduced with the `dict` below.

```
my_dict = {
'Circle': {'Ellipse',
           'Shape',
           'ShapeDriver'},

'Ellipse': {'Shape'},

'Rectangle': {'Shape'},

'Shape': {'Circle'},

'ShapeDriver': {'Circle',
                'Ellipse',
                'Shape'},

'Square': {'Rectangle'}
}
```

Steps to recreate:
1. create new `dict`
2. type `my_dict[`
3. press <tab> key to use tab-completion
4. crash

```
~ % bpython
bpython version 0.21 on top of Python 3.8.2 /Library/Developer/CommandLineTools/usr/bin/python3
>>> my_dict = {
... 'a/Circle': {'Ellipse',
...            'Shape',
...            'ShapeDriver'},
... 
... 'a/Ellipse': {'Shape'},
... 
... 'a/Rectangle': {'Shape'},
... 
... 'a/Shape': {'Circle'},
... 
... 'a/ShapeDriver': {'Circle',
...                 'Ellipse',
...                 'Shape'},
... 
... 'a/Square': {'Rectangle'}
... }
>>> my_dict['b/banana']
  File "/Users/arian/Library/Python/3.8/bin/bpython", line 8, in <module>
    sys.exit(main())
  File "/Users/arian/Library/Python/3.8/lib/python/site-packages/bpython/curtsies.py", line 201, in main
    exit_value = repl.mainloop(True, paste)
  File "/Users/arian/Library/Python/3.8/lib/python/site-packages/bpython/curtsies.py", line 121, in mainloop
    self.process_event_and_paint(e)
  File "/Users/arian/Library/Python/3.8/lib/python/site-packages/bpython/curtsies.py", line 85, in process_event_and_paint
    self.process_event(e)
  File "/Users/arian/Library/Python/3.8/lib/python/site-packages/bpython/curtsiesfrontend/repl.py", line 615, in process_event
    return self.process_key_event(e)
  File "/Users/arian/Library/Python/3.8/lib/python/site-packages/bpython/curtsiesfrontend/repl.py", line 757, in process_key_event
    self.on_tab()
  File "/Users/arian/Library/Python/3.8/lib/python/site-packages/bpython/curtsiesfrontend/repl.py", line 891, in on_tab
    cursor_and_line = self.matches_iter.substitute_cseq()
  File "/Users/arian/Library/Python/3.8/lib/python/site-packages/bpython/repl.py", line 297, in substitute_cseq
    self.update(
  File "/Users/arian/Library/Python/3.8/lib/python/site-packages/bpython/repl.py", line 318, in update
    self.start, self.end, self.current_word = self.completer.locate(
TypeError: cannot unpack non-iterable NoneType object
```

Image below so you can see the colors as well:
image

arian-deimling:
Fixes #917

### DIFF
--- a/bpython/line.py
+++ b/bpython/line.py
@@ -65,41 +65,14 @@ r"""  # bytes repr() begins with `b` character
     # end matching at closing double-quote if one is present
     "?"""
 
-_match_int_and_float = \
-r"""\-?         # match a negative sign if present
-    \d+         # match any number of digits
-    (?:\.\d*)?  # if there is a decimal point, match it and subsequent digits
-    |\-         # if only a negative sign has been entered, match it
-"""
-
-# used to match possible elements in a tuple
-_match_tuple_element = (
-    f'(?:{_match_single_quote_str_bytes}|'  # match single quote str/bytes repr
-    f'{_match_double_quote_str_bytes}|'     # or double quote str/bytes repr
-    f'{_match_int_and_float})'              # or int or float repr
-)
-
-_match_tuple = \
-rf"""  # match open parenthesis
-    \(
-    # matches any number of tuple elements followed by `,` then ` `
-    (?:{_match_tuple_element},[ ])*
-    # followed by another element being typed or one element tuple i.e. `(1,)`
-    (?:{_match_tuple_element},?)?
-    # match close parentesis if present
-    \)?
-"""
-
 # match valid identifier name followed by `[` character
 _match_dict_before_key = r"""[\w_][\w0-9._]*\["""
 
 _current_dict_key_re = LazyReCompile(
-    f'{_match_dict_before_key}('
+    f'{_match_dict_before_key}((?:'
     f'{_match_single_quote_str_bytes}|'
     f'{_match_double_quote_str_bytes}|'
-    f'{_match_int_and_float}|'
-    f'{_match_tuple}|'
-    f'{_match_all_dict_keys}|)',
+    f'{_match_all_dict_keys}|)*)',
     re.VERBOSE
 )
 
@@ -116,12 +89,12 @@ def current_dict_key(cursor_offset: int, line: str) -> Optional[LinePart]:
 _capture_dict_name = r"""([\w_][\w0-9._]*)\["""
 
 _current_dict_re = LazyReCompile(
-    f'{_capture_dict_name}('
+    f'{_capture_dict_name}((?:'
     f'{_match_single_quote_str_bytes}|'
     f'{_match_double_quote_str_bytes}|'
-    f'{_match_int_and_float}|'
-    f'{_match_tuple}|'
-    f'{_match_all_dict_keys}|)',
+    # f'{_match_int_and_float}|'
+    # f'{_match_tuple}|'
+    f'{_match_all_dict_keys}|)*)',
     re.VERBOSE
 )
 
--- a/bpython/line.py
+++ b/bpython/line.py
@@ -6,6 +6,7 @@ word."""
 
 from itertools import chain
 from typing import Optional, NamedTuple
+import re
 
 from .lazyre import LazyReCompile
 
@@ -35,31 +36,71 @@ def current_word(cursor_offset: int, line: str) -> Optional[LinePart]:
 
 
 # pieces of regex to match repr() of several hashable built-in types
-_match_identifiers_and_functions = r"""[a-zA-Z_][a-zA-Z0-9_.(),\[\] ]*"""
-_match_single_quote_str_bytes = \
-    r"""b?'(?:\\(?:\\\\)*['"nabfrtvxuU]|[^'\\])*(?:\\\\)*\\?'?"""
-_match_double_quote_str_bytes = r"""b?"[^"]*"?"""
-_match_int_and_float = r"""\-?\d+(?:\.\d*)?|\-"""
-_match_str_bytes_int_float = \
-    f'(?:{_match_single_quote_str_bytes}|{_match_double_quote_str_bytes}|' \
-    f'{_match_int_and_float})'
-_match_tuple = r"""\((?:""" \
-    f'{_match_str_bytes_int_float}' \
-    r""", )*(?:""" \
-    f'{_match_str_bytes_int_float}' \
-    r"""\,|""" \
-    f'{_match_str_bytes_int_float}' \
-    r""")?\)?"""
+_match_all_dict_keys = r"""[^\]]*"""
 
-_current_dict_key_re_base = r"""[\w_][\w0-9._]*\["""
+# https://docs.python.org/3/reference/lexical_analysis.html#string-and-bytes-literals
+_match_single_quote_str_bytes = \
+r"""  # bytes repr() begins with `b` character; bytes and str begin with `'`
+    b?'             
+    # after an even number of `\`, next `\` and subsequent char are interpreted
+    # as an escape sequence; this handles `\'` in the string repr()
+    (?:\\(?:\\\\)*['"nabfrtvxuU]|
+    # or match any non-`\` and non-single-quote character (most of the string)
+    [^'\\])*
+    # any even number of `\`; ensures `\\'` is interpreted as the end of string
+    (?:\\\\)*
+    # matches hanging `\` if one is present
+    \\?
+    # end matching at closing single-quote if one is present
+    '?
+"""
+
+# bytes and str repr() only uses double quotes if the string contains 1 or more
+# `'` character and exactly 0 `"` characters
+_match_double_quote_str_bytes = \
+r"""  # bytes repr() begins with `b` character
+    b?"
+    # string continues until a `"` character is reached
+    [^"]*
+    # end matching at closing double-quote if one is present
+    "?"""
+
+_match_int_and_float = \
+r"""\-?         # match a negative sign if present
+    \d+         # match any number of digits
+    (?:\.\d*)?  # if there is a decimal point, match it and subsequent digits
+    |\-         # if only a negative sign has been entered, match it
+"""
+
+# used to match possible elements in a tuple
+_match_tuple_element = (
+    f'(?:{_match_single_quote_str_bytes}|'  # match single quote str/bytes repr
+    f'{_match_double_quote_str_bytes}|'     # or double quote str/bytes repr
+    f'{_match_int_and_float})'              # or int or float repr
+)
+
+_match_tuple = \
+rf"""  # match open parenthesis
+    \(
+    # matches any number of tuple elements followed by `,` then ` `
+    (?:{_match_tuple_element},[ ])*
+    # followed by another element being typed or one element tuple i.e. `(1,)`
+    (?:{_match_tuple_element},?)?
+    # match close parentesis if present
+    \)?
+"""
+
+# match valid identifier name followed by `[` character
+_match_dict_before_key = r"""[\w_][\w0-9._]*\["""
+
 _current_dict_key_re = LazyReCompile(
-    f'{_current_dict_key_re_base}('
-    f'{_match_identifiers_and_functions}|'
+    f'{_match_dict_before_key}('
     f'{_match_single_quote_str_bytes}|'
     f'{_match_double_quote_str_bytes}|'
     f'{_match_int_and_float}|'
     f'{_match_tuple}|'
-    ')'
+    f'{_match_all_dict_keys}|)',
+    re.VERBOSE
 )
 
 
@@ -71,15 +112,17 @@ def current_dict_key(cursor_offset: int, line: str) -> Optional[LinePart]:
     return None
 
 
-_current_dict_re_base = r"""([\w_][\w0-9._]*)\["""
+# capture valid identifier name if followed by `[` character
+_capture_dict_name = r"""([\w_][\w0-9._]*)\["""
+
 _current_dict_re = LazyReCompile(
-    f'{_current_dict_re_base}('
-    f'{_match_identifiers_and_functions}|'
+    f'{_capture_dict_name}('
     f'{_match_single_quote_str_bytes}|'
     f'{_match_double_quote_str_bytes}|'
     f'{_match_int_and_float}|'
     f'{_match_tuple}|'
-    ')'
+    f'{_match_all_dict_keys}|)',
+    re.VERBOSE
 )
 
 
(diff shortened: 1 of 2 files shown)


thomasballinger:
I tried this out and found that this more general pattern meant the key matching too much of the line, so tab completion of the second dictionary on a line no longer works. I think `.*` is too broad a pattern, could we restrict it further? Maybe we could allow anything if it started with a `'` or `"` and once that quote is matched, not allow more?



It would be nice to add some tests, both of the new functionality you're adding (backslashes in strings work!) and this regression (second dictionary on a line should still work) like https://github.com/bpython/bpython/blob/b66a29fde838f98a39c08ce1b94f6cc8b8c3e86d/bpython/test/test_line_properties.py#L171-L184
These tests use a DSL to describe the expected match when calling one of these functions:
https://github.com/bpython/bpython/blob/b66a29fde838f98a39c08ce1b94f6cc8b8c3e86d/bpython/test/test_line_properties.py#L77-L86

so a test for this regression might look like
```
    self.assertAccess("(asdf['abc'], asdf[<'ab>|") 
```

Thanks for looking into this!

thomasballinger:
@arian-deimling could you add tests to keep track of the different cases we've talked about?

This pattern doesn't match
```
>>> d = {'asdf': 123, 'a\'b"\\c': 234}
>>> d['a\'
```
but that didn't work before either, so I think your code is a winner!

You can run the tests by running `pytest` in the repo.

arian-deimling:
@thomasballinger If you don't mind, I'd like like to work on this a little bit more. It's more nuanced than I originally thought, but I think with a few more minor modifications, we could have the tab completion working for most different types of potential dict keys (a.k.a. hashable types - https://stackoverflow.com/questions/14535730/what-does-hashable-mean-in-python/44880799#44880799). I agree that there will probably have to be a few edge cases that we cannot handle.

And to answer your first question, yes, I will add some tests!

arian-deimling:
@thomasballinger Ok - ready for you to have another look with these two commits. This newest version should be able to tab-complete all `str`, `bytes`, and `int` keys, `float` keys besides special values (i.e. nan, inf, -inf), and `tuple` keys for tuples containing any combination of the aforementioned key possibilities.

Also, that testing stuff is pretty neat and straightforward. The changes pass all tests as well as 9 new ones. 

This may be off topic, but I think this test case is irrelevant:
https://github.com/bpython/bpython/blob/b66a29fde838f98a39c08ce1b94f6cc8b8c3e86d/bpython/test/test_line_properties.py#L176
because in the case where you have typed `some_dict[abc`, the program is not trying to match `abc` as a key, but rather as the beginning of a variable or function name.

thomasballinger:
> This may be off topic, but I think this test case is irrelevant because in the case where you have typed some_dict[abc, the program is not trying to match abc as a key, but rather as the beginning of a variable or function name.

I agree that that code path isn't useful when running bpython, and I am ambivalent about inverting the test and changing the behavior. I wrote out the following to think through this.

---

If we implemented fuzzy completion for dict keys (which we do have implemented in `AttrCompletion` and `GlobalCompletion`) then it would useful to match this, assuming we still want DictKeyCompletion to take precedence over expression completion in that situation. I don't plan to do that though, and I'm not sure we'd want to make it so fuzzy that quotes will be added for you. If the completer were every changed to return matches that included global variables that referred to dictionary keys we would need it... but no plans for that. If we decided to show all matches instead of filtering on prefix we'd want this, but that doesn't sound very useful.

The only reason to leave this test case is that always matching the item in the dict key position makes it easier to describe what `DictKeyCompletion.locate()` does since its more in the spirit of what locate() is supposed to do according to the `BaseCompletion` class:

https://github.com/bpython/bpython/blob/b66a29fde838f98a39c08ce1b94f6cc8b8c3e86d/bpython/autocomplete.py#L225-L226

A reason to change the behavior and explicitly not match anything other than hashable literals is that it would be a more efficient solution in some cases than [what I implemented](https://github.com/bpython/bpython/commit/58f3edfb26875450bbd71b909cb003f6b942bdab) for deciding which completer to use: if `DictKeyCompletion.locate()` returned None, we could bail out of DictKeyCompletion.matches() before needing to evaluate the dictionary expression, get keys, and see if the expression is a prefix of any keys (and it never is, since the only candidate keys are literals). However if we want to bail out early we could just check manually in `DictKeyCompletion.matches` instead of changing what `locate()` means.

https://github.com/bpython/bpython/blob/b66a29fde838f98a39c08ce1b94f6cc8b8c3e86d/bpython/autocomplete.py#L653-L668

For now, starting to type an identifier in the dict key position means we will not be able to complete a dictionary key. So returning None for something that could not possibly be a prefix of a valid key literal does make sense: there is a dict key, but it's not a completable one and this code is all about completion. So I'd be ok with this change, but I think it's ok as-is too.

bpython/line.py
```diff
@@ -34,7 +34,33 @@ def current_word(cursor_offset: int, line: str) -> Optional[LinePart]:
...
+    f'{_match_str_bytes_int_float}' \
+    r""", )*(?:""" \
+    f'{_match_str_bytes_int_float}' \
+    r"""\,|""" \
+    f'{_match_str_bytes_int_float}' \
+    r""")?\)?"""
+
+_current_dict_key_re_base = r"""[\w_][\w0-9._]*\["""
+_current_dict_key_re = LazyReCompile(
+    f'{_current_dict_key_re_base}('
```
thomasballinger, line 56: the "_base" part of _current_dict_key_re_base confuses me a little, _base because it's the beginning of the pattern? because the dict is the base of the expression?

thomasballinger (commented):
Thanks for taking the time to get this right, this looks much more precise than we had it. I learned a bunch from reading it.

Does matching more types like this expand what works for substitutions on tab autocomplete? I'd like to understand the behavior change to figure out what the more complex code is buying us.

bpython/line.py
```diff
@@ -34,7 +34,33 @@ def current_word(cursor_offset: int, line: str) -> Optional[LinePart]:
     return LinePart(start, end, word)
 
 
-_current_dict_key_re = LazyReCompile(r"""[\w_][\w0-9._]*\[([\w0-9._(), '"]*)""")
+# pieces of regex to match repr() of several hashable built-in types
+_match_identifiers_and_functions = r"""[a-zA-Z_][a-zA-Z0-9_.(),\[\] ]*"""
+_match_single_quote_str_bytes = \
+    r"""b?'(?:\\(?:\\\\)*['"nabfrtvxuU]|[^'\\])*(?:\\\\)*\\?'?"""
+_match_double_quote_str_bytes = r"""b?"[^"]*"?"""
```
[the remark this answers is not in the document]
arian-deimling: Yea, based on some research and just testing different strings, repr() will only surround a string in double-quotes if it contains at least one single quote char (`'`) and exactly zero double quote chars (`"`).

bpython/line.py
arian-deimling: Yea, I could use some direction on the name if you don't like that, but basically, `base` is the part that I did not change which is everything before the `[` character after which 'key matching mode' is entered.

bpython/line.py
```diff
@@ -34,7 +34,33 @@ def current_word(cursor_offset: int, line: str) -> Optional[LinePart]:
     return LinePart(start, end, word)
 
 
-_current_dict_key_re = LazyReCompile(r"""[\w_][\w0-9._]*\[([\w0-9._(), '"]*)""")
+# pieces of regex to match repr() of several hashable built-in types
+_match_identifiers_and_functions = r"""[a-zA-Z_][a-zA-Z0-9_.(),\[\] ]*"""
+_match_single_quote_str_bytes = \
+    r"""b?'(?:\\(?:\\\\)*['"nabfrtvxuU]|[^'\\])*(?:\\\\)*\\?'?"""
```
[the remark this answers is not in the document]
arian-deimling: Agreed. I will take care of that.

bpython/line.py
```diff
@@ -34,7 +34,33 @@ def current_word(cursor_offset: int, line: str) -> Optional[LinePart]:
     return LinePart(start, end, word)
 
 
-_current_dict_key_re = LazyReCompile(r"""[\w_][\w0-9._]*\[([\w0-9._(), '"]*)""")
+# pieces of regex to match repr() of several hashable built-in types
```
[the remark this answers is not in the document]
arian-deimling: Good idea! I'll update it to include that.. honestly, it's even a little bit difficult for me to read even though I only wrote it a day ago.

arian-deimling:
> Thanks for taking the time to get this right, this looks much more precise than we had it. I learned a bunch from reading it.
> 
> Does matching more types like this expand what works for substitutions on tab autocomplete? I'd like to understand the behavior change to figure out what the more complex code is buying us.

Yes, now you can tab-complete any of the types that I added a regex for, and the way I tested for this was to make a `dict` with those kinds of keys and then type `my_dict[` and then make sure that for every new character I typed in, the completion options were still displayed to the user.

Tuples are a special case because you cannot tab-complete them unless you type `my_dict[|]` with the pipe character representing the cursor location. I'm not sure exactly what the cause is, but it matches the behavior prior to my changes, so I think it's okay. One thing I noticed in these cases where you type the closing brace `[` is that when you use the tab completion to finish your key, bpython adds a closing brace `]`, so you are left with two `]`. We may want to open another issue for that; however, this behavior is also consistent with the main branch on the bpython repo.

bpython/line.py
thomasballinger: maybe `_current_dict_key_dict_re`, or maybe since all of these substrings are not indented to be used separately, a different name like `_current_dict_key_dict_part` — OH or to match your current convention of `_match_*` maybe just `_dict` or `_dict_before_match` or something. 🤞 there'll be a beautiful was to do this with re.VERBOSE, but if not just changing to something shorter is fine. I just don't want to force readers to know the history of the code in order to understand the variable name.

arian-deimling: Renamed to `_match_dict_before_key`.

arian-deimling:
@thomasballinger Okay.. so I think I lost sight of what issue I was trying to solve but I think this final commit is the one!

This comment is what I'm basing this last change on:
>The only reason to leave this test case is that always matching the item in the dict key position makes it easier to describe what DictKeyCompletion.locate() does since its more in the spirit of what locate() is supposed to do according to the BaseCompletion class

So the original issue was that we weren't matching everything that can possibly be in the key position, and the primary issue is strings because they can contain the `]` character which is the natural way to detect the end of the dict key. To fix this I realized, we should first capture any string in the key position, and outside of strings, we should capture anything but the `]` character. If we capture any number or combination of these using something like:
` (  (?: regex_to_match_any_string | regex_to_match_anything_but_close_square_bracket )*  )`
then we will capture the entire key. This has the added benefit of allowing text completion of for a multitude of other possible dict keys, such as `range`s, `complex` numbers among others that my older version did not capture.

There is one regression here, but I think it is worth the additional functionality. Current in `bpython`, if you test
`abcd[<'>|]` the test will pass, but with my new version the version of the test that would pass is `abcd[<'|]>`. This is because the `'` initiates string pattern matching and therefore the `]` is considered part of the string that is being typed in. This means that you cannot tab-complete a string if you've typed `my_dict[|]` and then start typing the string at the cursor position; however, this regression is offset by the fact that you can tab-complete a string that contains the `]` character.

Let me know what you think, but I think the solution to the regression is by making a change in the `locate` or `match` function down the line.

arian-deimling:
@thomasballinger Can you re-run the test? I fixed the code to pass the black linter test.

arian-deimling:
I didn’t mean to do that. Just clicked on the button to see what is does. Sorry

thomasballinger:
No problem! I want Sebastian to take a look, and he's busy today so it'll be another day at least FYI.

bpython/line.py
```diff
@@ -6,6 +6,7 @@
 
 from itertools import chain
 from typing import Optional, NamedTuple
+import re
```
sebastinas, line 9: Nitpick: we usually have `import` first, then `from ... import ...`.

bpython/test/test_line_properties.py
```diff
@@ -178,10 +178,19 @@ def test_simple(self):
...
         # TODO self.assertAccess('d[d[<12|>')
         self.assertAccess("d[<'a>|")
+        self.assertAccess("object.dict['a'bcd'], object.dict[<'abc>|")
+        self.assertAccess(r"object.dict[<'a\'\\\"\n\\'>|")
+        self.assertAccess("object.dict[<\"abc'>|")
+        self.assertAccess("object.dict[<(1, 'apple', 2.134>|]")
+        self.assertAccess("object.dict[<(1, 'apple', 2.134)>|]")
+        self.assertAccess("object.dict[<-1000>|")
+        self.assertAccess("object.dict[<-0.23948>|")
+        self.assertAccess("object.dict[<'\U0001ffff>|")
```
sebastinas, line 193: I think it would make sense to also add some tests for string keys containing `[` and `]` .

arian-deimling: I misinterpreted this and added a test where the user is typing a string dict key inside `[` and `]` and found an issue so I fixed that by modifying the regex. Also added some tests for strings that contain `[` and `]`.

bpython/line.py
arian-deimling: Done.

arian-deimling:
@sebastinas I think it's good to go now.

arian-deimling:
Turned off `black` globally because I don't like double quotes and forgot to reenable it for this @sebastinas. Good to go now. Sorry

thomasballinger:
Thanks for the summary! Looks good to me.

> This means that you cannot tab-complete a string if you've typed my_dict[|] and then start typing the string at the cursor position; however, this regression is offset by the fact that you can tab-complete a string that contains the ] character.

That is unfortunate, I bet it's more common to autocomplete inside [] than to want to complete keys with ] in them; but it's more correct and if we want to fix this we can do it explicitly with a workaround instead of an accident of the way the code works. But there are other similar edge cases here, it would be neat if e.g. `my_dict["abc|"]` completed the key `"abcdef"` but prefix matching isn't going to do that. We can fix those sometime separately.

PR #929, opened 2021-10-07
Mypy types for autocomplete.py, args.py, config.py and curtsies.py

thomasballinger:
`mypy` now checks the whole package... but only these 4 files have these pragmas which require functions to have type annotations: 
```
# mypy: disallow_untyped_calls=True
# mypy: disallow_untyped_defs=True
```

The way to run this locally is
> python3.9 -m mypy

but you need to install some things first:
> python3.9 -m pip install types-backports types-requests types-setuptools types-toml types-pygments numpy watchdog twisted

You may also need to upgrade curtsies if you haven't installed it fresh because I made the types a little more lenient.
> python3.9 -m pip install --upgrade curtsies

### DIFF
--- a/bpython/autocomplete.py
+++ b/bpython/autocomplete.py
@@ -66,7 +66,7 @@ class AutocompleteModes(Enum):
     FUZZY = "fuzzy"
 
     @classmethod
-    def from_string(cls, value: str) -> Union[Any, None]:
+    def from_string(cls, value: str) -> Optional[Any]:
         if value.upper() in cls.__members__:
             return cls.__members__[value.upper()]
         return None
@@ -209,7 +209,7 @@ def method_match_substring(word: str, size: int, text: str) -> bool:
     return text in word
 
 
-def method_match_fuzzy(word: str, size: int, text: str) -> Union[Match, None]:
+def method_match_fuzzy(word: str, size: int, text: str) -> Optional[Match]:
     s = r".*%s.*" % ".*".join(list(text))
     return re.search(s, word)
 
@@ -236,7 +236,7 @@ class BaseCompletionType:
     @abc.abstractmethod
     def matches(
         self, cursor_offset: int, line: str, **kwargs: Any
-    ) -> Union[Set[str], None]:
+    ) -> Optional[Set[str]]:
         """Returns a list of possible matches given a line and cursor, or None
         if this completion type isn't applicable.
 
@@ -255,7 +255,7 @@ class BaseCompletionType:
         raise NotImplementedError
 
     @abc.abstractmethod
-    def locate(self, cursor_offset: int, line: str) -> Union[LinePart, None]:
+    def locate(self, cursor_offset: int, line: str) -> Optional[LinePart]:
         """Returns a Linepart namedtuple instance or None given cursor and line
 
         A Linepart namedtuple contains a start, stop, and word. None is
@@ -299,7 +299,7 @@ class CumulativeCompleter(BaseCompletionType):
 
         super().__init__(True, mode)
 
-    def locate(self, cursor_offset: int, line: str) -> Union[LinePart, None]:
+    def locate(self, cursor_offset: int, line: str) -> Optional[LinePart]:
         for completer in self._completers:
             return_value = completer.locate(cursor_offset, line)
             if return_value is not None:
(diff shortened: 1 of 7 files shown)


bpython/autocomplete.py
```diff
@@ -265,27 +286,31 @@ def shown_before_tab(self) -> bool:
...
             raise ValueError(
                 "CumulativeCompleter requires at least one completer"
             )
-        self._completers = completers
+        self._completers: Sequence[BaseCompletionType] = completers
 
         super().__init__(True, mode)
 
-    def locate(self, current_offset: int, line: str) -> Union[None, NoReturn]:
+    def locate(self, cursor_offset: int, line: str) -> Union[LinePart, None]:
```
sebastinas, line 302: `Optional[LinePart]` would work for the return type too.

bpython/autocomplete.py
```diff
@@ -301,28 +326,36 @@ def matches(
...
 
-    def matches(self, cursor_offset, line, **kwargs):
+    def matches(
+        self, cursor_offset: int, line: str, **kwargs: Any
+    ) -> Union[None, Set]:
         return self.module_gatherer.complete(cursor_offset, line)
 
-    def locate(self, current_offset, line):
-        return lineparts.current_word(current_offset, line)
+    def locate(self, cursor_offset: int, line: str) -> Union[LinePart, None]:
```
sebastinas, line 342: As above.

bpython/autocomplete.py
```diff
@@ -337,10 +370,10 @@ def matches(self, cursor_offset, line, **kwargs) -> Union[None, set]:
             matches.add(filename)
         return matches
 
-    def locate(self, current_offset, line):
-        return lineparts.current_string(current_offset, line)
+    def locate(self, cursor_offset: int, line: str) -> Union[LinePart, None]:
```
sebastinas, line 373: As above

bpython/curtsiesfrontend/interaction.py
```diff
@@ -1,5 +1,6 @@
 import greenlet
 import time
+from typing import Optional
```
sebastinas, line 3: Unused

bpython/curtsiesfrontend/repl.py
```diff
@@ -55,6 +60,8 @@
 )
 from ..translations import _
 
+InputOrOutput = Union[Literal["input"], Literal["output"]]
```
sebastinas, line 63: I would prefer to avoid the runtime dependency on `typing_extensions`. `InputOrOutput` could be an enum instead.

bpython/test/test_inspection.py
```diff
@@ -2,6 +2,9 @@
 import sys
 import unittest
 
+from typing import Optional
+from types import ModuleType
```
sebastinas, line 6: Unused

stubs/cwcwidth.pyi
```diff
@@ -0,0 +1,4 @@
+# official typing seems to require a second arg
+def wcswidth(s: str, n=100000000): ...
+
+__version__: str
```
sebastinas, line 4: `cwcwidth` should have type support. If it's buggy, I'd prefer to fix `cwcwidth`.

bpython/curtsies.py
```diff
@@ -196,16 +245,20 @@ def curtsies_arguments(parser):
             with repl.window as win:
                 with repl:
                     repl.height, repl.width = win.t.height, win.t.width
-                    exit_value = repl.mainloop(True, paste)
+                    repl.mainloop(True, paste)
```
sebastinas, line 248: Is this a change in behavior?

bpython/autocomplete.py
thomasballinger: I got rid of all of these except a few
```
-> Union[str, curtsies.events.Event, None]:
```
in curtsies.py.

bpython/curtsiesfrontend/repl.py
thomasballinger: there's already an enum and it even says
> (use LineTypeTranslator.INPUT or LineTypeTranslator.OUTPUT to avoid typing these strings)
a line above one of these that it changed! I just didn't use it.

stubs/cwcwidth.pyi
thomasballinger: For now I've included the second argument which was defaulting to None, I'll look into updating the types for cwcwidth to make the second argument optional.

bpython/curtsies.py
thomasballinger: `FullCurtsiesRepl.mainloop()` never returns a non-None value, so this is not a change.

stubs/cwcwidth.pyi
sebastinas: I think it just needs a release. It's fixed in master.

PR #936, opened 2021-11-01
added type references to pager.py, mypy produced no errors

Frostmaine:
First time adding to a project on github, started off small. Would really appreciate feedback.

### DIFF
--- a/bpython/pager.py
+++ b/bpython/pager.py
@@ -30,9 +30,10 @@ import pydoc
 import subprocess
 import sys
 import shlex
+from typing import List
 
 
-def get_pager_command(default: str = "less -rf") -> list[str]:
+def get_pager_command(default: str = "less -rf") -> List[str]:
     command = shlex.split(os.environ.get("PAGER", default))
     return command
 
@@ -53,6 +54,7 @@ def page(data: str, use_internal: bool = False) -> None:
         curses.endwin()
         try:
             popen = subprocess.Popen(command, stdin=subprocess.PIPE)
+            assert popen.stdin is not None
             data_bytes = data.encode(sys.__stdout__.encoding, "replace")
             popen.stdin.write(data_bytes)
             popen.stdin.close()
--- a/bpython/pager.py
+++ b/bpython/pager.py
@@ -20,6 +20,8 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 # THE SOFTWARE.
 
+# mypy: disallow_untyped_defs=True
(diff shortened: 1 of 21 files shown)


bpython/pager.py
```diff
@@ -30,29 +32,29 @@
...
 
-def page(data, use_internal=False):
+def page(data: str, use_internal: bool = False) -> None:
     command = get_pager_command()
     if not command or use_internal:
         page_internal(data)
     else:
         curses.endwin()
         try:
             popen = subprocess.Popen(command, stdin=subprocess.PIPE)
```
thomasballinger, line 55: For me this needs a
```suggestion
            assert popen.stdin is not None
            popen = subprocess.Popen(command, stdin=subprocess.PIPE)
```
to silence mypy errors about stdin maybe being None.

thomasballinger (changes requested):
Thanks for working on this!

It looks like some builds are failing because `list` can't be subscripted, I added a note about why.

When I run mypy, I get a few errors

```
$ python -m mypy
bpython/pager.py:57: error: Item "None" of "Optional[IO[bytes]]" has no attribute "write"
bpython/pager.py:58: error: Item "None" of "Optional[IO[bytes]]" has no attribute "close"
bpython/_internal.py:7: error: Incompatible types in assignment (expression has type "Callable[[str, bool], None]", variable has type "Callable[[str], None]")
Found 3 errors in 2 files (checked 64 source files)
```

If you figure out how our environments are different let me know and maybe we can add it to some developer docs.

The bpython/_internal.py file is getting more strictly checked now because it uses `page` defined in pager.py. Apparently the real pydoc.pager doesn't take that second argument. I'd be fine with a `type: ignore` for that.

```
pydoc.pager = page  # type: ignore
```

thomasballinger:
If you aren't able to mypy to produce these errors locally, you can see them in the CI runs on this page.
image

Frostmaine:
Linter workflow did fail, however it seemed to only be because I modified _internal.py by adding the suggested # type: ignore to pydoc.pager = page.

sebastinas:
Indeed,  the `black` check failed. So please run `black` on `bpython/_internal.py` and commit the fix.

thomasballinger:
Thanks for adding these types and fixing these CI issues, this looks good to me. The pypy3 CI failure is failing on main right now too.

Frostmaine:
Glad to complete my first contribution to a project that I use daily.  I
hope to take the  things I learned about the development process and
continue to contribute in my own small way to this project.  Any feedback
on how I have conducted my interactions with you are welcome.  Thank you so
much for your time.

PR #947, opened 2022-01-04
WIP: Adding type hints to formatter.py

Ben-Reg:
Adding type hints to formatter.py.

A note on line 128.  `token = token.parent` causes some problems with mypy because `token` is type `_TokenType` while `token.parent` is `Optional[_TokenType]`

The easy solution that I used below was to type `tokensource` like so:
`tokensource: Iterable[MutableMapping[Any, str]]` 

Another approach would be something like this:
`t: Optional[_TokenType] = token`
`t = t.parent`
If we do it that way, we can type `tokensource` more accurately:
`tokensource: Iterable[MutableMapping[_TokenType, str]]`

### DIFF
--- a/bpython/formatter.py
+++ b/bpython/formatter.py
@@ -28,9 +28,10 @@
 # mypy: disallow_untyped_calls=True
 
 
-from typing import MutableMapping, Any, Iterable, TextIO
+from typing import MutableMapping, Iterable, TextIO, Optional
 from pygments.formatter import Formatter
 from pygments.token import (
+    _TokenType,
     Keyword,
     Name,
     Comment,
@@ -113,16 +114,20 @@ class BPythonFormatter(Formatter):
         super().__init__(**options)
 
     def format(
-        self, tokensource: Iterable[MutableMapping[Any, str]], outfile: TextIO
+        self,
+        tokensource: Iterable[MutableMapping[_TokenType, str]],
+        outfile: TextIO,
     ) -> None:
         o: str = ""
         for token, text in tokensource:
+            t: Optional[_TokenType] = token
             if text == "\n":
                 continue
 
-            while token not in self.f_strings:
-                token = token.parent
-            o += f"{self.f_strings[token]}\x03{text}\x04"
+            while t not in self.f_strings:
+                if t is not None:
+                    t = t.parent
+            o += f"{self.f_strings[t]}\x03{text}\x04"
         outfile.write(o.rstrip())
 
 
--- a/bpython/formatter.py
+++ b/bpython/formatter.py
@@ -101,9 +101,7 @@ class BPythonFormatter(Formatter):
     straightforward."""
 
     def __init__(
-        self,
-        color_scheme: MutableMapping[str, str],
-        **options: str
+        self, color_scheme: MutableMapping[str, str], **options: str
     ) -> None:
         self.f_strings = {}
         for k, v in theme_map.items():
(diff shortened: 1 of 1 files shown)


Ben-Reg:
New to this - should I reformat formatter.py with black and resubmit the code?

thomasballinger:
Looks like you got it, thank you! Sorry for the delay here.

> A note on line 128. token = token.parent causes some problems with mypy because token is type _TokenType while token.parent is Optional[_TokenType]

I like the second one better, that seems truer to the code as it's currently written.

Ben-Reg:
Thanks, @thomasballinger.  I think the only way to avoid using `Any` as a type is to use a temporary variable - the reason being that `token` _has_ to be `_TokenType` (because of the line `token = token.parent` ... `None` cannot equal `None.parent`) and `token.parent` _has_ to be `Optional[_TokenType]` because that method can return either a token or `None`.

At least, that's how it seems to me.  I'll try another commit with a temp variable and you can make the call.  I'm probably using too much of your time on the little details, but I appreciate the help!

bpython/formatter.py
```diff
@@ -106,15 +113,21 @@ def __init__(self, color_scheme, **options):
...
             if text == "\n":
                 continue
 
-            while token not in self.f_strings:
-                token = token.parent
-            o += f"{self.f_strings[token]}\x03{text}\x04"
+            while t not in self.f_strings:
+                if t is not None:
+                    t = t.parent
+            o += f"{self.f_strings[t]}\x03{text}\x04"
```
sebastinas, line 130: `self.f_strings[None]` would raise an exception. It would be safe to bail out if `t.parent` is `None`.

Ben-Reg: Ah, yes.  I'll take another shot.

Ben-Reg:
@thomasballinger, is there anything else I need to do with this PR at this point?

thomasballinger:
Sorry about the delay, pining with that comment was the perfect action. I'll look into this more this morning, right now when I run mypy locally I'm getting 

```
(bpython) tomb ((0ca23b1...)) bpython$ mypy
bpython/formatter.py:114: error: No overload variant of "__init__" of "Formatter" matches argument type "Dict[str, str]"
bpython/formatter.py:114: note: Possible overload variants:
bpython/formatter.py:114: note:     def __init__(self, *, encoding: None = ..., outencoding: None = ..., **options: Any) -> None
bpython/formatter.py:114: note:     def __init__(self, *, encoding: str, outencoding: None = ..., **options: Any) -> None
bpython/formatter.py:114: note:     def __init__(self, *, encoding: None = ..., outencoding: str, **options: Any) -> None
```
which I'm sure it just my problem since it works in CI. Once I get this sorted (no action needed from you) I'll merge this.

thomasballinger:
I have to change the typing of `**options` to Any (which means Dict[str, Any] when used with `**`), which matches Pygments typing, to make this work locally for me. But it works without this change in CI. I figured this meant that my local enviroment was different than CI and CI should win, but I get the same thing from a fresh checkout on a linux box.

```
diff --git a/bpython/formatter.py b/bpython/formatter.py
index 20573bd..f216f21 100644
--- a/bpython/formatter.py
+++ b/bpython/formatter.py
@@ -28,7 +28,7 @@
 # mypy: disallow_untyped_calls=True


-from typing import MutableMapping, Iterable, TextIO
+from typing import Any, MutableMapping, Iterable, TextIO
 from pygments.formatter import Formatter
 from pygments.token import (
     _TokenType,
@@ -102,7 +102,7 @@ class BPythonFormatter(Formatter):
     straightforward."""

     def __init__(
-        self, color_scheme: MutableMapping[str, str], **options: str
+        self, color_scheme: MutableMapping[str, str], **options: Any
     ) -> None:
         self.f_strings = {}
         for k, v in theme_map.items():
```

I could imagine a reason for this: according to the types, it's not safe to include both `encoding` and `outencoding` with string values. Since I can do this with a fresh checkout, I think there's something wrong with our CI that we should fix, but in the meantime we can just make this change. I'll add this commit and merge.

thomasballinger:
Merged, thank you @Ben-Reg!

Issue #892, opened 2021-04-06
Add type annotations


sebastinas:
Some parts of the code are already annotated with types. Let's add to the remaining parts as well.

Mayur619:
Hi @sebastinas, I would like to work on this

sebastinas:
@Mayur619 Feel free to start with any file that does not already have type annotations. Easiest is probably to start with a module that only imports stuff from the standard library. For others,  we probably need stubs or type annotations in our dependencies.

supremestdoggo:
Hi! Can I help?

victoroliveirab:
Hey, @sebastinas, are there any more files needing help?

sebastinas:
@victoroliveirab  Yes, there are. You can start by running `mypy` on the `bpython` source tree and start from there.

Ben-Reg:
Hi - I'm new and this looks like a good place to start.  Can I start by working on formatter.py?

thomasballinger:
@Ben-Reg that sounds good, thanks for taking a shot at this!

Ben-Reg:
@thomasballinger, if I have questions, is this the best place to post them?  One of the variables in formatter.py is used in a way that makes assigning a type difficult.  I'm unsure whether I should rewrite it to allow for stricter typing, or just use the Any type.

thomasballinger:
@Ben-Reg opening an in-progress pull request might be the best way to ask since the context will be right there, but you can ask here too.

Which variable is it? Even though the file is short I'm having trouble guessing, so I can't suggest something specific. Starting with any sounds good to me, and we can discuss more as code review.

Ben-Reg:
Thanks @thomasballinger, I opened pull request #947 to add the context.  The issue is with the line `token = token.parent`

spotted-cat:
Heya I would like to help!? @sebastinas

thomasballinger:
@spotted-cat Great! The process is described above, do you have any questions about it? Go right ahead, looking forward to a pull request.

RCReddyN:
Hi! I would like to help.

PR #949, opened 2022-01-14
Added type annotations

RCReddyN:
I tried adding annotations in the following files, addressing issue #892 :
1. bpython/_internal.py
2. bpython/curtsiesfrontend/interpreter.py

@thomasballinger Please, look into it.

### DIFF
--- a/bpdb/__init__.py
+++ b/bpdb/__init__.py
@@ -106,9 +106,7 @@ def main():
             print("\t" + " ".join(sys.argv[1:]))
         except SystemExit:
             # In most cases SystemExit does not warrant a post-mortem session.
-            print(
-                "The program exited via sys.exit(). Exit status: ",
-            )
+            print("The program exited via sys.exit(). Exit status: ",)
             print(sys.exc_info()[1])
         except:
             traceback.print_exc()
--- a/bpython/_internal.py
+++ b/bpython/_internal.py
@@ -21,7 +21,7 @@ class _Helper:
             "or help(object) for help about object."
         )
 
-    def __call__(self, *args, **kwargs) -> None:
+    def __call__(self, *args: str, **kwargs: int) -> None:
         self.helper(*args, **kwargs)
 
 
--- a/bpython/args.py
+++ b/bpython/args.py
@@ -61,10 +61,7 @@ class RaisingArgumentParser(argparse.ArgumentParser):
 
 def version_banner(base: str = "bpython") -> str:
     return _("{} version {} on top of Python {} {}").format(
-        base,
-        __version__,
-        sys.version.split()[0],
-        sys.executable,
+        base, __version__, sys.version.split()[0], sys.executable,
     )
 
 
@@ -152,9 +149,7 @@ def parse(
         help=_("Set log level for logging"),
     )
     parser.add_argument(
-        "--log-output",
-        "-L",
-        help=_("Log output file"),
+        "--log-output", "-L", help=_("Log output file"),
     )
 
     if extras is not None:
--- a/bpython/curtsiesfrontend/interpreter.py
+++ b/bpython/curtsiesfrontend/interpreter.py
@@ -1,11 +1,11 @@
 import sys
-from typing import Any, Dict
-
-from pygments.token import Generic, Token, Keyword, Name, Comment, String
+from typing import Any, AnyStr, Dict, List, NoReturn, TextIO, IO, Tuple, Union
+from curtsies.formatstring import fmtstr, FmtStr
+from pygments.token import Generic, Token, Keyword, Name, Comment, String  # type: ignore
 from pygments.token import Error, Literal, Number, Operator, Punctuation
 from pygments.token import Whitespace
-from pygments.formatter import Formatter
-from pygments.lexers import get_lexer_by_name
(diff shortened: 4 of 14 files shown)


thomasballinger:
@RCReddyN could you fix these conflicts, and then we can run mypy in CI on this.

bpdb/__init__.py
```diff
@@ -106,9 +106,7 @@ def main():
             print("\t" + " ".join(sys.argv[1:]))
         except SystemExit:
             # In most cases SystemExit does not warrant a post-mortem session.
-            print(
-                "The program exited via sys.exit(). Exit status: ",
-            )
+            print("The program exited via sys.exit(). Exit status: ",)
```
thomasballinger, line 109: These formatting changes aren't desired, hopefully running black will fix them? Maybe you have a different autoformatter or different settings?

bpython/_internal.py
```diff
@@ -21,7 +21,7 @@ def __repr__(self) -> str:
             "or help(object) for help about object."
         )
 
-    def __call__(self, *args, **kwargs) -> None:
+    def __call__(self, *args: str, **kwargs: int) -> None:
```
thomasballinger, line 24: why are kwargs typed as int, is this based on the call signature of a helper? Maybe we should change the signature to match the helper object.

bpython/args.py
```diff
@@ -152,9 +149,7 @@ def callback(group):
         help=_("Set log level for logging"),
     )
     parser.add_argument(
-        "--log-output",
-        "-L",
-        help=_("Log output file"),
+        "--log-output", "-L", help=_("Log output file"),
```
thomasballinger, line 152: this looks like another bad reformat, I won't comment on any others but these formatting-only changes shouldn't be in this PR.

bpython/curtsiesfrontend/interpreter.py
```diff
@@ -1,11 +1,11 @@
...
-
-from pygments.token import Generic, Token, Keyword, Name, Comment, String
+from typing import Any, AnyStr, Dict, List, NoReturn, TextIO, IO, Tuple, Union
+from curtsies.formatstring import fmtstr, FmtStr
+from pygments.token import Generic, Token, Keyword, Name, Comment, String  # type: ignore
 from pygments.token import Error, Literal, Number, Operator, Punctuation
 from pygments.token import Whitespace
-from pygments.formatter import Formatter
-from pygments.lexers import get_lexer_by_name
+from pygments.formatter import Formatter  # type: ignore
```
thomasballinger, line 7: If you install the pygments types (see https://github.com/bpython/bpython/blob/main/.github/workflows/lint.yaml#L42) you won't need to ignore these

bpython/curtsiesfrontend/interpreter.py
```diff
@@ -43,13 +43,13 @@ class BPythonFormatter(Formatter):
     See the Pygments source for more info; it's pretty
     straightforward."""
 
-    def __init__(self, color_scheme, **options):
+    def __init__(self, color_scheme: Dict, **options: int) -> None:
```
thomasballinger, line 46: Dict isn't as helpful as `Dict[whatever, whatever]`

thomasballinger: Any why int here?

bpython/curtsiesfrontend/interpreter.py
```diff
@@ -43,13 +43,13 @@ class BPythonFormatter(Formatter):
...
         self.f_strings = {}
         for k, v in color_scheme.items():
             self.f_strings[k] = f"\x01{v}"
         super().__init__(**options)
 
-    def format(self, tokensource, outfile):
+    def format(self, tokensource: Any, outfile: Any) -> None:
```
thomasballinger, line 52: This kind of annotation isn't very useful, if you're not sure how to find out (feel free to ask!) then this might a good file to skip

thomasballinger (changes requested):

bpython/curtsiesfrontend/interpreter.py
```diff
@@ -60,7 +60,9 @@ def format(self, tokensource, outfile):
 
 
 class Interp(ReplInterpreter):
-    def __init__(self, locals: Dict[str, Any] = None, encoding=None):
+    def __init__(
+        self, locals: Dict[str, Any] = None, encoding: str = None
```
sebastinas, line 64: This should actually be marked with `Optional`.

bpython/curtsiesfrontend/interpreter.py
```diff
@@ -79,13 +81,13 @@ def write(err_line):
         self.write = write  # type: ignore
         self.outfile = self
 
-    def writetb(self, lines):
+    def writetb(self, lines: List[str]) -> None:
```
sebastinas, line 84: This could be `Iterable[str]`.

sebastinas (changes requested):

bpython/curtsiesfrontend/interpreter.py
sebastinas: … ah, but this is just noise from reformatting. This change will need to be rebased.

sebastinas:
@RCReddyN What's the status of this PR?

PR #954, opened 2022-03-10
WIP: cli.py type hints

Ben-Reg:
Disclaimer: Tackling this was a bit more ambitious than I thought it would be.  I tried to take my time and do research, but I'm assuming there will be some work still needed before it can be merged.

I'm in no hurry if we need to let this sit until you have time to look at it.  And of course, if you take a look and decide it just needs too much help, feel free to reject and I'll look for a less involved file to practice on before coming back to this one.

Notes:

- The `forward_if_not_current()` function wrapper gave me some trouble, but I think it's done in the recommended way.  A second opinion would be great.
- Also had trouble knowing how to type FakeDict, FakeStream and FakeStdin in a way that wouldn't cause errors
- I tried to add a short explanation in a comment whenever I had to use `type:ignore`

Thanks.

### DIFF
--- a/bpython/curtsiesfrontend/repl.py
+++ b/bpython/curtsiesfrontend/repl.py
@@ -13,7 +13,17 @@ import time
 import unicodedata
 from enum import Enum
 from types import TracebackType
-from typing import Dict, Any, List, Optional, Tuple, Union, cast, Type, TYPE_CHECKING
+from typing import (
+    Dict,
+    Any,
+    List,
+    Optional,
+    Tuple,
+    Union,
+    cast,
+    Type,
+    TYPE_CHECKING,
+)
 from .._typing_compat import Literal
 
 if TYPE_CHECKING:
@@ -362,7 +372,7 @@ class BaseRepl(Repl):
         logger.debug("starting parent init")
         # interp is a subclass of repl.Interpreter, so it definitely,
         # implements the methods of Interpreter!
-        super().__init__(cast('Interpreter', interp), config)
+        super().__init__(cast("Interpreter", interp), config)
 
         self.formatter = BPythonFormatter(config.color_scheme)
 
--- a/bpython/repl.py
+++ b/bpython/repl.py
@@ -430,7 +430,7 @@ class Repl:
         self.funcprops = None
         self.arg_pos: Union[str, int, None] = None
         self.current_func = None
-        self.highlighted_paren = None
+        self.highlighted_paren: Optional[Tuple[Any, List[Tuple[_TokenType, str]]]] = None
         self._C: Dict[str, int] = {}
         self.prev_block_finished: int = 0
         self.interact = Interaction(self.config)
@@ -649,7 +649,7 @@ class Repl:
         current name in the current input line. Throw `SourceNotFound` if the
         source cannot be found."""
 
-        obj = self.current_func
+        obj: Optional[Callable] = self.current_func
         try:
             if obj is None:
                 line = self.current_line
@@ -657,7 +657,8 @@ class Repl:
                     raise SourceNotFound(_("Nothing to get source of"))
                 if inspection.is_eval_safe_name(line):
                     obj = self.get_object(line)
-            return inspect.getsource(obj)
+            # Ignoring the next mypy error because we want this to fail if obj is None
+            return inspect.getsource(obj)  # type:ignore[arg-type]
         except (AttributeError, NameError) as e:
             msg = _("Cannot get source: %s") % (e,)
         except OSError as e:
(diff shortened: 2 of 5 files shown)


Ben-Reg:
Just looking at the errors that came up from the automated tests.

- The black error seems to be from a different file - I assume it's OK to ignore?
- I'll take a closer look at the mypy errors, but most of them seem to be coming from imports from other untyped files.  I ignored these when doing my own tests by using `--follow-imports=error`
- I'll follow up with the codespell error to see what that's about

thomasballinger:
Glancing at this, good work! I can take a longer look by Saturday, feel free to ask any questions before then though. For black, even if it isn't introduced in your PR go ahead and run black on that file to fix. If you install the dev dependencies. For mypy, we do actually want to fix these, the easiest time to add make sure consumers of this code are typed is is when you change the types of this file (and you might actually change the way you type this file to make these match). I'd recommend installing mypy and running it locally with the same flags we do in the CI command you're looking at.

It looks like codespell is just a spellchecker that found a word it thinks is mispelled.

Ben-Reg:
Ok, sounds good.  I've started working through the mypy import errors.  There's a lot from repl.py that will need to be typed, so it may take a little while to finish.

Ben-Reg:
Ok, I've made some progress, but ran into a few questions.  Mypy is pointing out a number of missing attributes in repl.py.  From what I can tell, it looks like it is assumed that these attributes should be provided by the subclass (and they are provided in cli.py).

I wasn't sure of how you would want me to handle this - do I create empty class variables and methods in repl.py to satisfy mypy?

Looks like the tests are highlighting some other issues as well.  I'll look into those.

Ben-Reg:
Ok, I've made some progress, but ran into a few questions.  Mypy is pointing out a number of missing attributes in repl.py.  From what I can tell, it looks like it is assumed that these attributes should be provided by the subclass (and they are provided in cli.py).

I wasn't sure of how you would want me to handle this - do I create empty class variables and methods in repl.py to satisfy mypy?

Looks like the tests are highlighting some other issues as well.  I'll look into those.

bpython/repl.py
```diff
@@ -50,7 +61,8 @@
     have_pyperclip = False
 
 from . import autocomplete, inspection, simpleeval
-from .config import getpreferredencoding
+from .cli import Statusbar
```
sebastinas, line 64: No, you cannot do that here. This leads to circular imports.

bpython/repl.py
```diff
@@ -349,7 +365,7 @@ def clear(self) -> None:
 
 
 class Interaction:
-    def __init__(self, config, statusbar=None):
+    def __init__(self, config: Config, statusbar: Optional[Statusbar] = None):
```
sebastinas, line 368: Note that the `urwid` frontends passes a different `Statusbar`. This needs be solved differently, e.g. by moving to the respective frontends.

sebastinas (changes requested):

thomasballinger:
@Ben-Reg I've not had time to look at this yet, will try to dig in this weekend. Or you're up to look together sometime send me an email at me@ballingt.com, we could screenshare and talk through some of the tricky bits.

Ben-Reg:
That's a generous offer, @thomasballinger.  I'll send you an email.

bpython/repl.py
thomasballinger: @sebastinas what do you mean by moving to the respective frontends here?

Ben-Reg:
Well, it looks like I may not get all the way there tonight.  I was able to find and fix the two pytest errors that were coming up before (with help from @thomasballinger - thanks!), but now I seem to be having problems with the python 3.7 tests (3.8 tests are running just fine on my machine).

Looks like using `curses.window` doesn't work with 3.7

bpython/curtsiesfrontend/repl.py
```diff
@@ -357,7 +370,9 @@ def __init__(
         )
         self.edit_keys = edit_keys.mapping_with_config(config, key_dispatch)
         logger.debug("starting parent init")
-        super().__init__(interp, config)
+        # interp is a subclass of repl.Interpreter, so it definitely,
+        # implements the methods of Interpreter!
+        super().__init__(cast("Interpreter", interp), config)
```
sebastinas, line 376: So why not mark `interp` to be of type `repl.Interpreter`?

bpython/repl.py
sebastinas: `Interaction.statusbar` is never used in `bpython/repl.py`. That's an implementation detail of the different repls that leaked into the base repl. If a repl needs that statusbar, they can add the statusbar to their `Interaction`.

bpython/repl.py
Ben-Reg: Thanks - this is fixed now.

bpython/repl.py
Ben-Reg: @sebastinas, I'm not positive this is the direction you wanted me to go, but I modified this to import Statusbar from both `cli` and `urwid` and then typed the parameter like this: `statusbar: Union["cliStatusbar", "urwidStatusbar", None] = None'

Let me know if I'm getting warmer or colder. :)

bpython/curtsiesfrontend/repl.py
Ben-Reg: @sebastinas, I've reworked this to use `repl.Interpreter` instead.

sebastinas:
Thanks, merged.

Issue #955, opened 2022-03-29
sys command results in error


thevibingcat:
when i enter the following code, this happens:
```>>> import sys
>>> v = sys.stdin.readline()
here is some text
>>> print(v)
here is some text
>>> v = sys.stdin.readline(13)
here is some text
[error text blah blah blah]```


what is supposed to happen is sys is supposed to save the first 13 characters (including spaces) of "here is some text".
if i do this in the actual IDLE, it does what is intended. but this does not happen in bpython. please fix this and any other commands that do not match up with the original python.
I am using Ubuntu 20.04.1, and the IDLE i mentioned is python 3.8 and bpython 0.18
Thank you.

sebastinas:
This is now fixed thanks to @ulisesojeda.

PR #960, opened 2022-05-27
Allowing sys.stind.readline receive size parameter

ulisesojeda:
Issue: https://github.com/bpython/bpython/issues/955
 
- Reproducing sys.stdin.readline behavior for size parameter.

### DIFF
--- a/bpython/curtsiesfrontend/repl.py
+++ b/bpython/curtsiesfrontend/repl.py
@@ -179,7 +179,7 @@ class FakeStdin:
         )
         self.cursor_offset += 1
 
-    def readline(self, size: int = -1) -> str:
+    def readline(self, size: int = -1) -> Union[str, Any]:
         if not isinstance(size, int):
             raise TypeError(
                 f"'{type(size).__name__}' object cannot be interpreted as an integer"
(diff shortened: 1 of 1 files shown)


bpython/curtsiesfrontend/repl.py
```diff
@@ -179,12 +179,19 @@ def add_input_character(self, e: str) -> None:
         )
         self.cursor_offset += 1
 
-    def readline(self):
-        self.has_focus = True
-        self.repl.send_to_stdin(self.current_line)
-        value = self.coderunner.request_from_main_context()
-        self.readline_results.append(value)
-        return value
+    def readline(self, size=-1):
```
sebastinas, line 182: While at it, could you also add type annotations here? That would be `def readline(self, size: int =  -1) -> str`.

bpython/curtsiesfrontend/repl.py
```diff
@@ -179,12 +179,19 @@ def add_input_character(self, e: str) -> None:
...
+        if not isinstance(size, int):
+            raise TypeError(
+                f"'{type(size).__name__}' object cannot be interpreted as an integer"
+            )
+        elif size == 0:
+            return ""
+        else:
```
sebastinas, line 189: No need for the else here - without it, the diff would contain less noise.

sebastinas (changes requested):

sebastinas (approved):

PR #961, opened 2022-06-09
Added type annotations to curtsiesfrontend/interpreter.py: BPythonFormatter.format()

amuletcoin:
Finished adding type hints to curtsiesfrontend/interpreter.py's functions, specifically the BPythonFormatter class's format() method.


sebastinas:
Sorry, miss clicked.

The black regression is easy to fix afterwards, but the mypy regressions needs to be fixed before this can go forward.

PR #967, opened 2022-08-29
Modify the requirements to integrate rich with BPython

JesFragoso:


sebastinas:
I don't see any clear motivation why this needs to be integrated in bpython directly. If you want to use `rich`, put into your `PYTHONSTARTUP` file.

PR #975, opened 2022-11-04
Fix watchdog auto-reloading

thomasballinger:
The module loading process has changed a lot (dropping support for 3.4 helps here), now auto-reloading works for me again.


sebastinas:
@thomasballinger I'm not seeing a difference. How was auto-reloading supposed to work?

thomasballinger:
Import a module from the current director (or elsewhere), press F5 and then modify the file: the session should re-run (like F6) with the new version of the file.

sebastinas (approved):

PR #992, opened 2023-07-09
Fix __signature__ support if object has a __file__

gpotter2:
Followup to https://github.com/bpython/bpython/pull/932. I missed an edge-case :(

If a class uses exotic `__signature__` typings, but also has a `__file__` (that `inspect.getsourcelines` finds), then it is ignored whenever `_fix_default_values` doesn't work. In this case it should just 'trust' the `__signature__`.


sebastinas:
Could you please add a test case covering such a class?

gpotter2:
I updated the previous test to test this edge case :p

PR #996, opened 2023-07-22
Improve autocompletion of parameters

gpotter2:
Two minor improvements to the completion:

1. auto-completion for parameters is done without needing to pass the beggining of a parameter. To give an example:

```python
class Test:
    def __init__(self, aaa, aab, aac):
        pass
```
**with the PR**: this allows to do: `Test(` then TAB directly to select the arguments
image

**upstream/main**: you need to do `Test(a` before completion kicks in. `Test(` gives nothing
image

2. auto-completion results are ordered slightly differently: parameters should be prioritized when available.

**this PR:**
image

**upstream/main**:
image

### DIFF
--- a/bpython/autocomplete.py
+++ b/bpython/autocomplete.py
@@ -604,7 +604,12 @@ class ParameterNameCompletion(BaseCompletionType):
         return matches if matches else None
 
     def locate(self, cursor_offset: int, line: str) -> Optional[LinePart]:
-        return lineparts.current_word(cursor_offset, line)
+        r = lineparts.current_word(cursor_offset, line)
+        if r and r.word[-1] == "(":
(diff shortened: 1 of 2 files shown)


bpython/autocomplete.py
```diff
@@ -742,6 +747,16 @@ def get_completer(
             double underscore methods like __len__ in method signatures
     """
 
+    def _cmpl_sort(x: str) -> Tuple[Any, ...]:
```
sebastinas, line 750: Could be `Tuple[bool, str]` instead.

gpotter2:
```suggestion
    def _cmpl_sort(x: str) -> Tuple[bool, ...]:
```

PR #1015, opened 2024-04-25
Update `setup.py` for more up-to-date/robust customisation of the `build` command

### LINKED ISSUE #1014: Customising `build.sub_commands` directly from `distutils` has been deprecated and is about to be removed
The following feature in setuptools has been deprecated for almost 2 years and is about to be removed:

https://github.com/pypa/setuptools/blob/1ed759173983656734c3606e9c97a348895e5e0c/setuptools/command/build.py#L13-L27

It might be a good idea to import `build` directly from setuptools for the following code:

https://github.com/bpython/bpython/blob/ca13bf5727d91d3208e84297c2829aa8474b4620/setup.py#L131-L139

(`build` is available directly from setuptools, starting on version v62.4.0)

abravalheri:
Closes #1014.

This PR proposes 2 changes:

- Import `build` from setuptools (motivation in #1014)
- Avoid patching of a class attribute of a class that owned by the project (`build.sub_commands`).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [build-system]
 requires = [
-  "setuptools >= 43",
+  "setuptools >= 62.4.0",
 ]
 build-backend = "setuptools.build_meta"
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,4 +4,4 @@ cwcwidth
 greenlet
 pyxdg
 requests
-setuptools
+setuptools>=62.4.0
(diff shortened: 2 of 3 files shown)


abravalheri:
> ModuleNotFoundError: No module named 'setuptools.command.build'

Maybe this can be fixed by running `python -m build` instead of `python setup.py build` (as that would enforce the build requirement already specified in `pyproject.toml`).

PR #1019, opened 2024-06-01
Add type annotations

sumangaire52:


sebastinas:
The change to `Never` requires a fallback for older Python versions.

sumangaire52:
> The change to `Never` requires a fallback for older Python versions.

I have reverted back to NoReturn for now.

PR #1031, opened 2025-01-28
Add short project description

homeworkprod:
Should be picked up by PyPI and others.

Currently, bpython appears in PyPI search results with a description of "None".


homeworkprod:
The test and codespell errors appear to be unrelated to this change.

